PR #2, opened 2019-01-07
Hide 2FA link in org and hide non functional 2FA buttons

BlackDex:
Added CSS to hide non functional 2FA buttons and hide the 2FA link in the organization interface.


dani-garcia:
Thanks!

PR #6, opened 2019-05-10
Fixed Org Billing Form

BlackDex:


dani-garcia:
I've looked into this and the patch only partially worked, I've made some changes in the latest commit that should solve the issue. Thanks anyway!

PR #9, opened 2019-12-25
Rework

ypid:


dani-garcia:
Thanks, makes sense! As you could see I'm not much of a shell expert

PR #10, opened 2020-02-20
pack into subdirectory

### LINKED ISSUE #7: repack into own directory
Currently the vault directory does not have a leading directory. May I kindly request packing it with a predictable name? This makes packaging easier and prevents surprises when people do not `tar tf` the tarball first.

So the traditional output would look like this:
```
% tar tf bw_web_v2.12.0.tar.gz|head -n 5
bw_web_v2.12.0/app/
bw_web_v2.12.0/app/main.553962d815b08b8f918b.css
bw_web_v2.12.0/app/main.553962d815b08b8f918b.css.map
bw_web_v2.12.0/app/main.553962d815b08b8f918b.js
bw_web_v2.12.0/app/main.553962d815b08b8f918b.js.map
```

Thanks!

mqus:
This fixes #7

### DIFF
--- a/package_web_vault.sh
+++ b/package_web_vault.sh
@@ -76,10 +76,9 @@ npm run dist
 #find build -name "*.map" -delete
 
 # Prepare the final archives
-cd build
-tar -czvf "../../$OUTPUT_NAME" ./* --owner=0 --group=0
+tar -czvf "../$OUTPUT_NAME" build --owner=0 --group=0
 
-cd ../..
+cd ..
 
 if [[ $UPLOAD_VAULT =~ ^[Yy]$ ]]
 then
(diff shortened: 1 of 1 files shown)


dani-garcia:
Sounds good to me, but maybe we should have the folder name be `web-vault`, as that's what bitwarden_rs expects by default.

package_web_vault.sh
```diff
@@ -76,10 +76,9 @@ npm run dist
 #find build -name "*.map" -delete
 
 # Prepare the final archives
-cd build
-tar -czvf "../../$OUTPUT_NAME" ./* --owner=0 --group=0
+tar -czvf "../$OUTPUT_NAME" build --owner=0 --group=0
```
mqus: > Sounds good to me, but maybe we should have the folder name be `web-vault`, as that's what bitwarden_rs expects by default.

like that?
```suggestion
mv build web-vault
tar -czvf "../$OUTPUT_NAME" web-vault --owner=0 --group=0
```

dani-garcia: Yeah that works!

PR #11, opened 2020-02-22
Add web vault support for alternate base dir (subdir/subpath) hosting

jjlin:
Instead of assuming the base URL is always `window.location.origin`,
infer it from `window.location.href`.

Requires dani-garcia/bitwarden_rs#868 for matching backend support.


dani-garcia:
Sounds good to me, thanks! I'll merge this one before and update the web vault for the docker images before merging the other PR

PR #18, opened 2020-09-22
Added support for v2.16.1 and small css fix.

BlackDex:
- Updated patch support for v2.16.1
- Hidden an extra form for taxes (not hidden in v2.16.0 either)

bitwarden/web commit-hash: 949f61f1a46a0ce680e9a39e6240f7b66b1b6efe

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,8 +22,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/web/commit/$VAULT_VERSION
 #
-# Using https://github.com/bitwarden/web/releases/tag/v2.16.0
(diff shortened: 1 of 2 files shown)


BlackDex:
Maybe it is better to just rename this to v2.16.0, if so, please let me know.

dani-garcia:
Well if it does apply to 2.16.0 then it probably makes more sense to update that one instead of a new one, but it doesn't matter much becasue i won't create a new build for 2.16.0 now that a newer release exists.

Can you also update the VAULT_VERSION in the Dockerfile with the commit hash?

BlackDex:
Sure, ill do that right now

BlackDex:
@dani-garcia: Done

PR #22, opened 2021-01-24
Update to upstream v2.18.0 release

jjlin:


patches/v2.18.0.patch
```diff
@@ -0,0 +1,154 @@
+diff --git a/package-lock.json b/package-lock.json
+index 437d9cad..f5e4a5c2 100644
+--- a/package-lock.json
++++ b/package-lock.json
+@@ -1,6 +1,6 @@
+ {
+   "name": "bitwarden-web",
+-  "version": "2.17.1",
++  "version": "2.18.0",
```
dani-garcia, line 9: Does this version change in the package lock affect anything?

jjlin: I don't think so, but it's not necessary either as it gets changed during `npm install`. I'll remove this part, thanks for noticing.

PR #30, opened 2021-02-20
Added warning when no HTTPS during register.

BlackDex:
We already had a patch to warn users they need HTTPS when they login.
This was not working for the register form.

- Added a warning during the submit of the register form.
- Changed the check to see if `crypto.subtle` is defined or not.
- Removed the `webpack.config.js` patch, since this now is in upstream.
- Updated README.md shields and small comment regarding how to create
  the patch because of changes to the submodule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,21 @@
 # Bitwarden_rs web vault builds
 
 [![GitHub Release](https://img.shields.io/github/release/dani-garcia/bw_web_builds.svg)](https://github.com/dani-garcia/bw_web_builds/releases/latest)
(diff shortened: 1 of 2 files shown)


BlackDex:
> Applying the patch give me this
> error: jslib/src/angular/components/register.component.ts: no folder or file of this type

Then you probably forgot the pull in the submodules.

```bash
git clone --recurse-submodules https://github.com/bitwarden/web.git web-vault
```

Also see:
- https://github.com/dani-garcia/bw_web_builds/blob/9b68e33580e09d92d7a82dfb7484a17c76e648f1/Dockerfile#L28
- https://github.com/dani-garcia/bw_web_builds/blob/9b68e33580e09d92d7a82dfb7484a17c76e648f1/package_web_vault.sh#L25

PR #37, opened 2021-05-12
Preliminary changes for v2.20.0

jjlin:
v2.20.0 hasn't been released yet, but there are some significant changes.
This patch is missing changes to `src/app/settings/two-factor-u2f.component.ts`;
this file doesn't exist anymore since upstream moved from U2F to its successor,
WebAuthn, which Vaultwarden doesn't have backend support for yet.

### DIFF
--- a/patches/v2.20.0.patch
+++ b/patches/v2.20.0.patch
@@ -1,0 +1,237 @@
+Submodule jslib contains modified content
+diff --git a/jslib/src/angular/components/register.component.ts b/jslib/src/angular/components/register.component.ts
+index ae45add..e6a5590 100644
+--- a/jslib/src/angular/components/register.component.ts
++++ b/jslib/src/angular/components/register.component.ts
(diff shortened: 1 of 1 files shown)


patches/v2.20.0.patch
```diff
@@ -0,0 +1,237 @@
...
+     }
+ 
+     isDev(): boolean {
+-        return process.env.ENV === 'development';
++        return false;
+     }
+ 
+     isSelfHost(): boolean {
+-        return process.env.SELF_HOST.toString() === 'true';
++        return true;
```
dani-garcia, line 234: This needs to be false, otherwise it will require uploading a license file to create an organization

jjlin: Ah, okay, I didn't notice that. There's a new trash message that depends on the self-hosting flag. I'll do it a different way then.

jjlin:
v2.20.0/v2.20.1 were released earlier today anyway. Closing in favor of #38.

PR #41, opened 2021-06-30
v2.21.0 Patch

idanoo:
Unsure if any extra changes are required, I have just mapped the previous patch to the new filenames/line numbers.


feskolech (approved):

BlackDex:
Thanks for the update.

We have to update some stuff on the vaultwarden server first before we can use v2.21.0.

PR #42, opened 2021-07-15
Enable emergency contact

thelittlefireman:


dani-garcia:
Included in https://github.com/dani-garcia/bw_web_builds/commit/4414f436648633cadfcc005fbb3031d478f47b3f with the update to v2.23.0. Thanks!

PR #45, opened 2021-08-30
Updated web-vault to latest v2.22.2 version.

BlackDex:
There needs to be a small change on the vaultwarden server for this to
be fully supported as far as i could tell.

So maybe not yet ready for merge/tagging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,8 +25,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/web/commit/$VAULT_VERSION
 #
-# Using https://github.com/bitwarden/web/releases/tag/v2.21.1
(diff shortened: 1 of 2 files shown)


BlackDex:
Just did a quick check, it all seems to work fine.
Only one item which is triggered only when we have emergency access merged, which isn't the case yet, so this can be used as the main web-vault. It has some fixes regarding performance improvements with large vaults according to the logs.

So, it's merge ready after all.

Rainson12:
i think there is an issue with hiding the tax info on the orgs page. It should be 
```
app-org-account > div.secondary-header:nth-child(7) { @extend %bwrs-hide; }
app-org-account > div.secondary-header:nth-child(7) + p { @extend %bwrs-hide; }
app-org-account > div.secondary-header:nth-child(7) + p + form { @extend %bwrs-hide; }
```
so index 7 instead of 3?

PR #68, opened 2022-06-02
Update web-vault to v2022.05.0 and misc changes

BlackDex:
- Updated web-vault to v2022.05.0
- Made several changes to the build scripts for ease of building/patching etc...
  You can now run `make checkout` to pull/update the repo to the wanted version
  There is also a `make patch-web-vault` to patch the repo
  A `make generate-patch` to create a patch file from the current changes
  And many more changes
- Also fixed an issue reported here: https://github.com/dani-garcia/vaultwarden/issues/2506


dani-garcia:
Nice! If now the makefile is the main entrypoint, maybe we could move the scripts into a separate scripts folder?

BlackDex:
Done :)

PR #72, opened 2022-06-29
Update web-vault to v2022.6.0

BlackDex:
Updated the web-vault to the latest released version.
This version is also using the new repo `bitwarden/clients` instead of the old `bitwarden/web`.
All files have been changed where needed.


dani-garcia:
When using package-web-vault.sh script, you now have to use `web-v2022.6.0`, if you use `v2022.6.0` like before the script fails to find the git tag, should we add that prefix ourselves instead?

PR #81, opened 2022-09-26
[WIP] feat: Bump vault to v2022.09.02

GeekCornerGH:
Feel free to leave a review
- Also catched a bug with 404 page, not the cleanest catch, tell me if you have suggestions

### DIFF
--- a/patches/v2022.9.2.patch
+++ b/patches/v2022.9.2.patch
@@ -1,0 +1,302 @@
+diff --git a/apps/web/src/404.html b/apps/web/src/404.html
+index 6cf5e36..54d1149 100644
+--- a/apps/web/src/404.html
++++ b/apps/web/src/404.html
+@@ -42,11 +42,10 @@
+         </a>
+       </p>
+       <p>
+-        You can <a href="/">return to the web vault</a>, check our
+-        <a href="https://status.bitwarden.com/">status page</a> or
+-        <a href="https://bitwarden.com/contact/">contact us</a>.
++        You can <a href="/">return to the web vault</a>, or
++        <a href="https://github.com/dani-garcia/vaultwarden">contact us</a>.
+       </p>
+     </div>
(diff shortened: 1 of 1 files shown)


stefan0xC:
Why has the patch file been added as a binary file? :raised_eyebrow: 

Regarding your changes to `404.html`: does your vaultwarden run in a subfolder?

GeekCornerGH:
> Why has the patch file been added as a binary file? :raised_eyebrow: 
> 
Uhh why did this happend :thinking:, will look at this today 


> Regarding your changes to `404.html`: does your vaultwarden run in a subfolder?
Yes, I know in nested subfolders it will still break however, do you have a better way to go to the vault itself?

stefan0xC:
> Yes, I know in nested subfolders it will still break however, do you have a better way to go to the vault itself?

Not really. It might be neccessary to use something like a [`sub_filter`](https://nginx.org/en/docs/http/ngx_http_sub_module.html) in nginx (or a [`Substitute`](https://httpd.apache.org/docs/2.4/mod/mod_substitute.html) directive in apache) to replace `href="/` with `href="/subfolder/` and `src="/` with `src="/subfolder/` in order to have everything working correctly for the 404 page.

BlackDex:
This is going to be difficult regarding the 404 page.
I think the only viable solution would be to embed this within Vaultwarden it self and have a custom Vaultwarden error page, or embed this specific html and use Handlebars to provide the correct path, same as we do for the admin interface.

Maybe we can add the needed variable here in the patches, by adding `{{urlpath}}` in front of all the needed URL's, and somehow add the 404.html file from the web-vault to Handlebars and parse it (one time during startup) and show it.

GeekCornerGH:
@BlackDex Regarding Commit hash, I will try to fix it tomorrow. 
Regarding 404, I will revert the changes.
And about register, I probably won't be able to help you. I'd like to learn rust at some point, but I'm enough busy right now with my Node.JS projects 🤣

GeekCornerGH:
Closing in favor of #82

PR #82, opened 2022-10-07
update web-vault to v2022.9.2

stefan0xC:
Depends on dani-garcia/vaultwarden#2803 because the registerComponent in v2022.9.2 requires a JsonResult.

### DIFF
--- a/patches/v2022.9.2.patch
+++ b/patches/v2022.9.2.patch
@@ -1,0 +1,301 @@
+diff --git a/apps/web/src/404.html b/apps/web/src/404.html
+index 6cf5e363d..54d11495c 100644
+--- a/apps/web/src/404.html
++++ b/apps/web/src/404.html
+@@ -42,11 +42,10 @@
+         </a>
+       </p>
+       <p>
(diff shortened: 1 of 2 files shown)


patches/v2022.9.2.patch
```diff
@@ -0,0 +1,19169 @@
...
++      );
++      return;
++    }
++
+     let email = this.formGroup.get("email")?.value;
+     email = email.trim().toLowerCase();
+     let name = this.formGroup.get("name")?.value;
+diff --git a/package-lock.json b/package-lock.json
+index d59e4fdcb..2156bf50d 100644
+--- a/package-lock.json
```
jjlin, line 304: The patch shouldn't include diffs for `package-lock.json`

jjlin (changes requested):

patches/v2022.9.2.patch
stefan0xC: ah, thanks.

BlackDex (approved):

PR #87, opened 2022-10-19
Update README.md

tessus:
- fix the links to bitwarden web-vault
- 8bit Solutions LLC. -> Bitwarden, Inc. (to be consistent with https://github.com/dani-garcia/vaultwarden/blob/main/README.md)


dani-garcia:
Thanks!

Issue #85, opened 2022-10-16
2022.10.0 cuts off "Send" menu items


tessus:
The last version cuts off the menu items for Send:

<img width="338" alt="image" src="https://user-images.githubusercontent.com/223439/196056753-e3459875-54f7-4af8-bd9a-3ad2949d5faa.png">

All Sen...
Te..
Fi..

This makes no sense, since there's more than enough space.

BlackDex:
Seems to work just fine on my browser.
Have you tried an cognition/private tab or other browser?
It could be some extension maybe, or maybe the zoom level.

tessus:
I tried with 2 browsers (one w/o extensions) and also Private Browsing on both.

Unfortunately I still experience the same issue. Also, no matter the zoom level, it's always the same.

tessus:
Ha, I was able to narrow it down. It does not happen with Safari 14.1.2. Only with Firefox 105.0.3 (64-bit) and Firefox Developer Edition 106.0b9 (64-bit).

It also works on Chrome 88.0.4324.192 (yes, I know it's an ancient version, since I don't use Chrome and locked it to that version).

Apparently it's a problem with the Firefox engine.

BlackDex:
This also seems to happen on `vault.bitwarden.com`, so, this is not something we can fix on our side.
It seems to have been reported upstream already here: https://github.com/bitwarden/clients/issues/3789
And also fixed via here: https://github.com/bitwarden/clients/pull/3791

So, we just have to wait until they release a new version of the web-vault, because backporting only this fix and creating a new release is a bit to much i think 😄 .

But, thanks for reporting!

tessus:
Thanks for the info.

PR #88, opened 2022-10-19
update web-vault to v2022.10.1

stefan0xC:
a small release which only includes the fix for #85

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,8 +28,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2022.10.0
-ARG VAULT_VERSION=feb10128f0963bf6a9dabf23d1a01d8ae8ea117c
+# Using https://github.com/bitwarden/clients/releases/tag/web-v2022.10.1
+ARG VAULT_VERSION=ae435905f7758262694a78fca54b799b5e7be0a5
 
 RUN git clone https://github.com/bitwarden/clients.git /vault
 WORKDIR /vault


Dockerfile
```diff
@@ -28,8 +28,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2022.10.0
-ARG VAULT_VERSION=feb10128f0963bf6a9dabf23d1a01d8ae8ea117c
+# Using https://github.com/bitwarden/clients/releases/tag/web-v2022.10.1
+ARG VAULT_VERSION=ae435905f7758262694a78fca54b799b5e7be0a5
```
tessus, line 32: Isn't the commit rather `02ae7ad9b3eb01051295e8a2ca627cb4773b424a`? At least in the master branch?

The commit `ae435905f7758262694a78fca54b799b5e7be0a5` is part of [hotfix-rc-web](https://github.com/bitwarden/clients/compare/hotfix-rc-web). I am just wondering what happens, if that branch is gone.

stefan0xC: ae435905f7758262694a78fca54b799b5e7be0a5 is simply the commit hash of the release [web-v2022.10.1 ](https://github.com/bitwarden/clients/releases/tag/web-v2022.10.1).

tessus: A git hash is only created for a commit (and also for a merge commit, and squashing will also create new git hashes). Tagging does not create a commit hash, but adds a tag to a commit.

It seems that the tag was added to a commit in the branch `hotfix-rc-web`. However, that tag does not seem to exist in the master branch.

I am just saying that if someone were to delete the branch `hotfix-rc-web`, you could not reference this commit point with the git hash that is in that branch, because it would be gone.

But maybe I completely misunderstood how git works.

BlackDex: As far as i know the hashes from the tagges keep existing. Someone could overwrite a tag if course and then it could probably fall.

Also, if you take the commit hash from master, then you will also get a bunch of other changes which should not be part of that specific release, like new upcoming features for example, because the commit chain there is different.

So, this is the correct hash 😃.

@stefan0xC could you also make sure that `DuckDuckGo` is hidden from the `Forwarded Alias` list? Same as is done for FireFox already. It doesn't work for self-hosted and only generates errors and issue reports.

stefan0xC: @BlackDex I made a separate PR to address this issue.

BlackDex (approved):

PR #91, opened 2022-10-24
update web-vault to v2022.10.2

stefan0xC:
this release contains a bug fix for exporting organization vault data as .json (encrypted) using password protection

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,8 +28,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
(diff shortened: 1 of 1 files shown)


GeekCornerGH (changes requested):
Probably still need to duplicate the patch file right?

tessus:
@GeekCornerGH nope, it will take the last one that is available for the `2022.10.0` series, which is [v2022.10.0.patch](https://github.com/dani-garcia/bw_web_builds/blob/master/patches/v2022.10.0.patch)

GeekCornerGH:
Didn't knew this, good to know

GeekCornerGH (approved):

PR #94, opened 2022-11-02
don't stash the node_modules folder

stefan0xC:
I was wondering why `make checkout` always takes so long and it's because we are using `git stash --all` which stashes also ignored files, i.d. the `node_modules` folder, too. Not sure if this is necessary, since we use `npm ci` to build the web-vault so we don't really need to stash it, because it will always [start with a clean build](https://docs.npmjs.com/cli/v8/commands/npm-ci) anyway.

### DIFF
--- a/scripts/checkout_web_vault.sh
+++ b/scripts/checkout_web_vault.sh
@@ -35,7 +35,7 @@ else
     # If there already is a checked-out repo, lets clean it up first.
     pushd "${VAULT_FOLDER}"
         # Stash current changes if there are any, we don't want to loose our work if we had some
-        git stash --all --quiet &> /dev/null || true
+        git stash --include-untracked --quiet &> /dev/null || true
         # Checkout the master repo first
         git checkout master
         git reset --hard


BlackDex:
I actually never use the host provide tools, and always use the docker commands. That way my host keeps clean and those files never appear.

Not sure if this will work fully. I'll need to check.

tessus:
Also, there's no need to check out the entire history. A shallow clone should be more than enough.

tessus:
sorry, I missed the fact that at that point the cloning already happened. So in that case in the other `if` branch then...

stefan0xC:
> Also, there's no need to check out the entire history. A shallow clone should be more than enough.

Maybe in the `Dockerfile` here:
https://github.com/dani-garcia/bw_web_builds/blob/613870ebe67fe803a1e380d526b20f7861eb489f/Dockerfile#L34

tessus:
That too, bu5t I meant actually here:

https://github.com/dani-garcia/bw_web_builds/blob/a21f082591bb207bf54e2bd6ae98c2ca9827ae16/scripts/checkout_web_vault.sh#L33

BlackDex:
A shallow clone cloud work, though, since there is going to be executed a `git fetch --all --tags` it still downloads a lot of data.
But, it might save a few MB's when the initial `git clone` has just a depth of `1` though, i have not tried it actually.

If it saves a lot, then it might be useful. Else it will just pull in afterwards, or cause issues with `git checkout`.

tessus:
No worries, it was just an idea, since this PR wanted to address the time it took to checkout. Shallow clones save time and space. ;-)

Come to think of it, shallow clones only work well when you need the git branch HEAD. Since this is not the case here, please disregard my comment.

It's too cumberscane to use `--shallow-since`. Since it's thrown out anyway and done most likelt via some sort of a a pipeline, no need to bother that much.

Sorry for the noise.

stefan0xC:
You can do something like
 ```
RUN git clone --depth 1 --single-branch --branch web-v2022.10.2 -c advice.detachedHead=false https://github.com/bitwarden/clients.git /vault
```

But as I said, it makes more sense for the docker build as you don't keep the repository anyway. (Since I already have the whole repository checked out it would not matter to me anyway.)

stefan0xC:
I've made a separate #95 to keep the discussion on topic.

BlackDex (approved):
I think this is ok.
Since we do not add custom files by default this should be ok.
If we were to add custom files, then this will cause an issue though.

PR #95, opened 2022-11-02
make only a shallow clone

stefan0xC:
In a docker build we don't keep the repository so it does not make sense to checkout the whole history, a shallow clone should suffice.

When updating the web-vault version we'd have to change the branch name too because it's not easily possible to checkout just a specific commit, I think.


BlackDex:
This is not really intuitive. It also breaks the docker ARG which you could use instead.

stefan0xC:
> This is not really intuitive. It also breaks the docker ARG which you could use instead.

Do you know how?

stefan0xC:
Instead of the `git clone` we could do the following
```
RUN git init
RUN git remote add origin https://github.com/bitwarden/clients.git
RUN git fetch --depth 1 origin "${VAULT_VERSION}"
RUN git -c advice.detachedHead=false checkout "${VAULT_VERSION}"
```

BlackDex:
I don't think it is possible with a single command.
Also, we want to use a specific commit instead of tags, just to be sure.

Update: And i see you may have a solution here.
That could be an option, we just need to make sure it works fully.

Dockerfile
```diff
@@ -31,9 +31,10 @@ USER node
 # Using https://github.com/bitwarden/clients/releases/tag/web-v2022.10.2
 ARG VAULT_VERSION=703860f1ac7bad2dfa9a98640a568db26fdcf5ed
 
-RUN git clone https://github.com/bitwarden/clients.git /vault
 WORKDIR /vault
-
+RUN git init
```
jjlin, line 35: It would probably be cleaner to just do something like
```Dockerfile
ARG VAULT_TAG=web-v2022.10.2
ARG VAULT_HASH=703860f1ac7bad2dfa9a98640a568db26fdcf5ed

RUN git clone --depth 1 --branch ${VAULT_TAG} https://github.com/bitwarden/clients.git /vault
WORKDIR /vault
RUN test "$(git rev-parse HEAD)" = ${VAULT_HASH}
```

jjlin (changes requested):

Dockerfile
stefan0xC: Good to know that we don't need `--single-branch`. But with an added `-c advice.detachedHead=false` the line will become maybe too long again? But as far as I've tested it, this works just as well, thanks for the suggestion.

Since the `apply_patches.sh` cannot use the current `VAULT_VERSION` to find a patch file I think it's fine not to define this variable either. But maybe we want to add the following https://github.com/dani-garcia/bw_web_builds/blob/613870ebe67fe803a1e380d526b20f7861eb489f/scripts/generate_patch_file.sh#L18-L22 to to the `apply_patches.sh` script too, not sure if there would be a downside?

BlackDex: I do not fully agree with this.
A Tag can be overwritten, a commit hash is fixed.
That could cause a different hash. Also, it means you need to provide two separate values if you want to pull/clone instead of just one commit hash.

I'm still in favor of just one variable for selecting the correct version.

jjlin: It's not clear to me exactly what you're referring to, but if you mean my use of two variables, `VAULT_TAG` and `VAULT_HASH`, I think this is better because the build process will verify that these two values correspond as expected (currently, `VAULT_TAG` is only recorded in a comment).

Hypothetically, if Bitwarden were to tag a commit, and then the next day, they decide that that commit had a bug and they retag a different commit with the same tag, then we would continue using the old commit with no indication that there is an issue (until GitHub removes the old commit, if that even happens). Of course, retagging with the same tag instead of using a new one is bad practice, especially for a public repo, but it's something that has been known to happen.

BlackDex: It could also mean that the patch isn't valid for the new tagged vault. Using a hash is more safe in these cases.

I do not see an issue with the solution @stefan0xC provide.

stefan0xC: @jjlin I think https://github.com/dani-garcia/bw_web_builds/blob/613870ebe67fe803a1e380d526b20f7861eb489f/Dockerfile#L13-L14 would not work for specific commit hashes anymore because [`--branch`](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---branchltnamegt) only works with tags and branches.
Not sure if this is actually used and I have not tested it but it should still work with my slightly less elegant multi-line approach.

stefan0xC: Okay, checking out a tag like `web-v2022.10.2` directly does not work.
```
error: pathspec 'web-v2022.10.2' did not match any file(s) known to git
Error: building at STEP "RUN git -c advice.detachedHead=false checkout "${VAULT_VERSION}"": while running runtime: exit status 1
```

Seems like we have to `git fetch --tags` too. :thinking:

BlackDex (approved):
Looks like this works just fine for the build process.

PR #99, opened 2022-12-02
make backup of previous web-vault

stefan0xC:
Currently, if a `builds/web-vault` directory exists (e.g. `builds/bw_web_v2022.11.1`) the script will copy the newly built web-vault to `builds/web-vault/web-vault`.

By renaming the folder we should be able to prevent this.

### DIFF
--- a/scripts/tar_web_vault.sh
+++ b/scripts/tar_web_vault.sh
@@ -18,13 +18,28 @@ mkdir -pv "${BASEDIR}/../${OUTPUT_FOLDER}"
 pushd "${VAULT_FOLDER}/apps/web"
 
 VAULT_VERSION=$(get_web_vault_version)
-OUTPUT_NAME="${OUTPUT_FOLDER}/bw_web_${VAULT_VERSION}"
+OUTPUT_NAME="${BASEDIR}/../${OUTPUT_FOLDER}/bw_web_${VAULT_VERSION}"
+DATE_FORMAT="${DATE_FORMAT:-%Y-%m-%dT%H:%M:%S%z}"
+
+# Preserve previous output
+if [[ -f "${OUTPUT_NAME}.tar.gz" ]];
+then
+    DATE_SUFFIX=$(date +"${DATE_FORMAT}" -r "${OUTPUT_NAME}.tar.gz")
+    mv "${OUTPUT_NAME}.tar.gz" "${OUTPUT_NAME}_${DATE_SUFFIX}.tar.gz"
+fi
+
+# Cleanup previous output directory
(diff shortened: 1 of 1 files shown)


scripts/tar_web_vault.sh
```diff
@@ -18,13 +18,21 @@ mkdir -pv "${BASEDIR}/../${OUTPUT_FOLDER}"
...
+OUTPUT_NAME="${BASEDIR}/../${OUTPUT_FOLDER}/bw_web_${VAULT_VERSION}"
+DATE_FORMAT="${DATE_FORMAT:-%Y-%m-%dT%H:%M:%S%z}"
 
 mv build web-vault
 # Tar the web-vault
-tar -czvf "${BASEDIR}/../${OUTPUT_NAME}.tar.gz" web-vault --owner=0 --group=0
+tar -czvf "${OUTPUT_NAME}.tar.gz" web-vault --owner=0 --group=0
+
+# Preserve previous output directory
+if [[ -e ${OUTPUT_NAME} ]];
```
BlackDex, line 29:
```suggestion
if [[ -d ${OUTPUT_NAME} ]];
```
I think using `-d` is better then `-e`, since it is more specific.

scripts/tar_web_vault.sh
```diff
@@ -18,13 +18,21 @@ mkdir -pv "${BASEDIR}/../${OUTPUT_FOLDER}"
...
+then
+    mv ${OUTPUT_NAME}{,-$(date +${DATE_FORMAT} -r ${OUTPUT_NAME})}
```
BlackDex, line 31: First, the arguments need to be quoted to prevent argument globing, else path's with spaces could cause issues.
Second, i think creating an archive of the previous output would be better, and after that remove it.
Also since `mv` has the same quirks as `cp` which is why you crated this PR in the first place.
This will only happen if you run this script twice or more right after each other, but it could happen.

```suggestion
    tar -czf "${OUTPUT_NAME}-$(date +"${DATE_FORMAT}" -r "${OUTPUT_NAME}").tar.gz" "${OUTPUT_NAME}" --owner=0 --group=0
    rm -rf "${OUTPUT_NAME}"
```

BlackDex (changes requested):

scripts/tar_web_vault.sh
stefan0xC: Thanks for the review. Since we already create an archive, I think we can preserve that file instead and remove the directory.

>  This will only happen if you run this script twice or more right after each other, but it could happen.

Yeah, it happens frequently to me while testing changes.

BlackDex (approved):

PR #100, opened 2022-12-02
improve 404 page and rename web vault pageTitle

stefan0xC:
Since the 404 page currently still has the Bitwarden logo, I've created a new one:

vaultwarden-404-error-page

And I was wondering, if we should call it Vaultwarden Web Vault now too?


BlackDex:
This is going to be too much for the patch file in my opinion.

I have thought about this already, i think it is better that we create a static page baked in into Vaultwarden or via a handlebars file and use that for the 404.

That way we can also make it work correctly for sub path.

BlackDex:
Looks like you seem to also changed the title. I looked for that, but didn't seem to be able to find it. Looks like i was searching for the wrong function hehe.

stefan0xC:
I'll make a new PR for changing only the page title.

PR #101, opened 2022-12-04
change pageTitle to Vaultwarden and add favicons

stefan0xC:
Not sure if this is okay but I've added custom favicons because it would have seemed weird to me to only change the pageTitle and still use the Bitwarden logo.


BlackDex:
Same for these png's, please use `oxipng` to shrink there size :)

stefan0xC:
> Same for these png's, please use `oxipng` to shrink there size :)

Oh, I thought I did. Must have forgotten to add the correct parameters. 😅

tessus:
@stefan0xC if you change everything from Bitwarden to Vaultwarden the login page should be changed as well:

image

tessus:
In that case wouldn't it make sense to upgrade to v2022.11.2 as well, otherwise the changes are in the tree, but no release is available for that fix.

stefan0xC:
> In that case wouldn't it make sense to upgrade to v2022.11.2 as well, otherwise the changes are in the tree, but no release is available for that fix.

I think that is a separate issue. I mean assuming there was no `v2022.11.2` release it should also be possible to make an updated `v2022.11.1` release of `bw_web_builds`. In this case a new `v2022.11.2` release should be easily possible as there are no changes to the patch required.

tessus:
> it should also be possible to make an updated v2022.11.1 release of bw_web_builds

Yes, this is true. But it seems to me that releases are not very often created (neither in this repo nor in the vaultwarden repo). This is not a complaint, but an observation. As I mentioned before, I think that testing a release is more important than creating a release for every few commits. Although it depends. If there's a critical bug or a compat issue, releases should be created right away (and they are in this project as far as I have seen).

Personally I couldn't care less, since I can create binaries or in this case the web vault myself. I can also merge PRs locally before doing so, thus for me releases are not a requirement, nor needed.

I just thought that the chances are higher to get a new release cut when there's also a bump in the version. That's all. ;-)

tessus:
I haven't tested this PR yet, but will it also use a black favicon for `/admin`?

Currently, a FF tab looks like this when browsing to `/admin`:

image

BlackDex:
> I haven't tested this PR yet, but will it also use a black favicon for `/admin`?
> 
> Currently, a FF tab looks like this when browsing to `/admin`:
> 
> image

These icon's have nothing to do with the admin.
And, for me, it looks ok 😛 , but I have a dark theme haha.

BlackDex:
@stefan0xC Regarding the `favicon-*` images, to me it looks a bit weird in the browsers as a favicon.
That is probably because there is no white border around the whole logo.

Maybe you could shrink the logo a bit, and have at least a 1 or 2 pixel white border.
Because now the logo touches the borders on all sides it looks a bit strange to me.

dani-garcia:
I think @BlackDex is right and the favicons need a little bit of padding to separate the black icon from the white background. But other than that this looks good.

Once this is merged we should also update the admin page favicon to be black-on-white as well to keep it consistent.

BlackDex:
@stefan0xC much better!

BlackDex (approved):

PR #105, opened 2022-12-14
update web-vault to v2022.12.0

tessus:
Upstream notes:

- Bug fixes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,8 +28,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2022.11.2
-ARG VAULT_VERSION=9c5aabcf21d932c7b60aad52e0e8ccb017d45e47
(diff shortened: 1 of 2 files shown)


GeekCornerGH (approved):

BlackDex (changes requested):
The patch file looks like an exact copy of `v2022.11.1.patch`.
While it does work, it isn't fully valid. There are some offsets different in the `v2022.12.0` patch.

Run the `make generate-patch` and force-push it so it will have a fully clean patch matching the correct offset. This could help with the next round of changes if the offset changes further.

tessus:
Done.

It's interesting though that the apply script did not mention it. I think it only said `patch xxx applied successfully`. Or I missed it, because it was just one occasion where the offset waws different:

```diff
-index 115d53401..e62638e35 100644
+index bb8bf7707..49f2444c8 100644
 --- a/apps/web/src/app/core/web-platform-utils.service.ts
 +++ b/apps/web/src/app/core/web-platform-utils.service.ts
-@@ -188,11 +188,11 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
+@@ -192,11 +192,11 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
    }
```

tessus:
@BlackDex for some reason it's not possibe to re-request a review.

PR #106, opened 2023-01-11
update web-vault to v2023.1.0

stefan0xC:
There has been a new release: [web-v2023.1.0](https://github.com/bitwarden/clients/releases/tag/web-v2023.1.0)

Also hide duckduckgo and firefox relay again from the username generator (as that has been changed in the recent releases).

### DIFF
--- a/patches/v2023.1.0.patch
+++ b/patches/v2023.1.0.patch
@@ -256,6 +256,19 @@ index 652ee24a1..5a63562cc 100644
          size: 160,
        });
      }, 100);
+diff --git a/apps/web/src/app/tools/generator.component.ts b/apps/web/src/app/tools/generator.component.ts
+index 6a5f00c19..1bf8daf31 100644
+--- a/apps/web/src/app/tools/generator.component.ts
++++ b/apps/web/src/app/tools/generator.component.ts
+@@ -40,7 +40,7 @@ export class GeneratorComponent extends BaseGeneratorComponent {
+       route,
+       window
+     );
+-    if (platformUtilsService.isSelfHost()) {
++    if (true) {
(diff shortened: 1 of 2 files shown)


patches/v2023.1.0.patch
```diff
@@ -0,0 +1,432 @@
...
++/* Hide Tax Info and Form in Organization settings */
++app-org-account > div.secondary-header:nth-child(3) { @extend %vw-hide; }
++app-org-account > div.secondary-header:nth-child(3) + p { @extend %vw-hide; }
++app-org-account > div.secondary-header:nth-child(3) + p + form { @extend %vw-hide; }
++
++/* Hide Firefox Relay - It is CORS protected for SaaS only */
++#forwardtype_firefoxrelay, label[for^=forwardtype_firefoxrelay] { @extend %vw-hide; }
++
++/* Hide DuckDuckGo - It does not work for self-hosted either. */
++#forwardtype_duckduckgo, label[for^=forwardtype_duckduckgo] { @extend %vw-hide; }
```
BlackDex, line 379: I suggest to remove these styles.
I think the rest should still be there as far as i can tell.

BlackDex (changes requested):

patches/v2023.1.0.patch
stefan0xC: done :+1:

BlackDex:
> As @GeekCornerGH pointed out in #108 there has been added a new Avatar Color Chooser so we need to implement the missing endpoint `/api/accounts/avatar` and handling in https://github.com/dani-garcia/vaultwarden to be able to save the new color.

That is indeed something to add. Though, it isn't a breaking feature in a sense to hide this via CSS, since it will only break when trying to change it.

BlackDex (approved):

PR #113, opened 2023-02-15
update web-vault to v2023.2.0

stefan0xC:
There has been a new web-vault release [web-v2023.2.0](https://github.com/bitwarden/clients/releases/tag/web-v2023.2.0)

From a first test run with vaultwarden's current main branch and a quick glance at the changes ([v2023.1.0..v2023.2.0](https://github.com/bitwarden/server/compare/v2023.1.0...v2023.2.0) and [web-v2023.1.1...web-v2023.2.0](https://github.com/bitwarden/clients/compare/web-v2023.1.1...web-v2023.2.0)) there seems to be at least a new `GET /organizations/<org_uuid>/collections/details` endpoint we need to implement in vaultwarden.

### DIFF
--- a/patches/v2023.2.0.patch
+++ b/patches/v2023.2.0.patch
@@ -1,0 +1,445 @@
+diff --git a/apps/web/src/404.html b/apps/web/src/404.html
+index 5a650b3673..275c34f726 100644
+--- a/apps/web/src/404.html
++++ b/apps/web/src/404.html
+@@ -28,7 +28,7 @@
+       <div class="container inner banner">
+         <div class="row align-items-center">
+           <div class="col brand">
+-            <i class="bwi bwi-shield"></i>&nbsp; <strong>bit</strong>warden
++            <i class="bwi bwi-shield"></i>&nbsp; <strong>Vault</strong>warden
+           </div>
+         </div>
+       </div>
+@@ -42,11 +42,10 @@
+         </a>
+       </p>
+       <p>
+-        You can <a href="/">return to the web vault</a>, check our
+-        <a href="https://status.bitwarden.com/">status page</a> or
+-        <a href="https://bitwarden.com/contact/">contact us</a>.
++        You can <a href="/">return to the web vault</a>, or
++        <a href="https://github.com/dani-garcia/vaultwarden">contact us</a>.
+       </p>
+     </div>
+-    <div class="container footer text-muted content">© Copyright 2023 Bitwarden, Inc.</div>
++    <div class="container footer text-muted content">Vaultwarden (unofficial Bitwarden&reg; server)</div>
+   </body>
+ </html>
+diff --git a/apps/web/src/app/app.component.ts b/apps/web/src/app/app.component.ts
+index b9dac0c6db..e7b84133eb 100644
+--- a/apps/web/src/app/app.component.ts
++++ b/apps/web/src/app/app.component.ts
+@@ -172,6 +172,10 @@ export class AppComponent implements OnDestroy, OnInit {
+             break;
+           }
+           case "showToast":
++            if (typeof message.text === "string" && typeof crypto.subtle === 'undefined') {
++              message.title="This browser requires HTTPS to use the web vault";
++              message.text="Check the Vaultwarden wiki for details on how to enable it";
++            }
+             this.showToast(message);
+             break;
+           case "setFullWidth":
+diff --git a/apps/web/src/app/core/init.service.ts b/apps/web/src/app/core/init.service.ts
+index 8baa7c0ddc..8f3b5d4641 100644
+--- a/apps/web/src/app/core/init.service.ts
++++ b/apps/web/src/app/core/init.service.ts
+@@ -37,11 +37,23 @@ export class InitService {
+   ) {}
+ 
+   init() {
++    function getBaseUrl() {
++      // If the base URL is `https://vaultwarden.example.com/base/path/`,
++      // `window.location.href` should have one of the following forms:
++      //
++      // - `https://vaultwarden.example.com/base/path/`
++      // - `https://vaultwarden.example.com/base/path/#/some/route[?queryParam=...]`
++      //
++      // We want to get to just `https://vaultwarden.example.com/base/path`.
++      let baseUrl = window.location.href;
++      baseUrl = baseUrl.replace(/#.*/, '');  // Strip off `#` and everything after.
++      baseUrl = baseUrl.replace(/\/+$/, ''); // Trim any trailing `/` chars.
++      return baseUrl;
++    }
+     return async () => {
+       await this.stateService.init();
+ 
+-      const urls = process.env.URLS as Urls;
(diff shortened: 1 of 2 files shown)


patches/v2023.2.0.patch
```diff
@@ -0,0 +1,423 @@
...
+@@ -77,11 +77,7 @@ export class VaultComponent implements OnInit, OnDestroy {
+   ) {}
+ 
+   async ngOnInit() {
+-    this.trashCleanupWarning = this.i18nService.t(
+-      this.platformUtilsService.isSelfHost()
+-        ? "trashCleanupWarningSelfHosted"
+-        : "trashCleanupWarning"
+-    );
++    this.trashCleanupWarning = this.i18nService.t("trashCleanupWarningSelfHosted");
```
BlackDex, line 189: I'm missing the same change for the `trashCleanupWarningSelfHosted` translation for the `apps/web/src/vault/app/vault/vault.component.ts` file.

BlackDex (changes requested):
I'm also missing a change to `libs/components/src/menu/menu.component.html b/libs/components/src/menu/menu.component.html` which adds `tw-overflow-x-auto`

```diff
diff --git a/libs/components/src/menu/menu.component.html b/libs/components/src/menu/menu.component.html
index 15f78351a..6d3421ffc 100644
--- a/libs/components/src/menu/menu.component.html
+++ b/libs/components/src/menu/menu.component.html
@@ -1,7 +1,7 @@
 <ng-template>
   <div
     (click)="closed.emit()"
-    class="tw-flex tw-shrink-0 tw-flex-col tw-rounded tw-border tw-border-solid tw-border-secondary-500 tw-bg-background tw-bg-clip-padding tw-py-2"
+    class="tw-flex tw-shrink-0 tw-flex-col tw-rounded tw-border tw-border-solid tw-border-secondary-500 tw-bg-background tw-bg-clip-padding tw-py-2 tw-overflow-x-auto"
     [attr.role]="ariaRole"
     [attr.aria-label]="ariaLabel"
     cdkTrapFocus
```

patches/v2023.2.0.patch
stefan0xC: Ah, sorry I think I was tired when checking the differences. I thik I've noticed that the second `trashCleanupWarningSelfHosted` change was missing but I did not notice that the change affected different files. Sorry about that. :facepalm:

I don't think `tw-overflow-x-auto` was missing though, or I am still not seeing it.

BlackDex: I'm not seeing the `tw-overflow-x-auto` change in your commit.

stefan0xC: > I'm not seeing the `tw-overflow-x-auto` change in your commit.

This change is at the bottom of the file (see [line 437](https://github.com/dani-garcia/bw_web_builds/pull/113/files#diff-a9baf2210c79a9297202cf14ded28a154e6174c670211872b6ae595d91abfe6dR437))? Or does it not get applied? I've just tested it to make sure and it seems to work for me so I am a bit confused.

tessus:
Hi stefan0xC. I also noticed a few commits in the bw server that have to do with something called SecretsManager. Unfortunately their PRs do not explain what that is, nor did I find anything in the bw documentation.

These changes are huge. I this something that must be implemented in vw as well?

BlackDex:
> Hi stefan0xC. I also noticed a few commits in the bw server that have to do with something called SecretsManager. Unfortunately their PRs do not explain what that is, nor did I find anything in the bw documentation.
> 
> These changes are huge. I this something that must be implemented in vw as well?

That is a Bitwarden licensed feature, so not something we can implement. It's kinda a hashicorp vault feature.

patches/v2023.2.0.patch
BlackDex: Ah, sorry, could have sworn it wasn't there in the first commit though, sorry for that.

patches/v2023.2.0.patch
```diff
@@ -0,0 +1,440 @@
...
+-    return process.env.NODE_ENV === "development";
++    return false;
+   }
+ 
+   isSelfHost(): boolean {
+-    return WebPlatformUtilsService.isSelfHost();
++    return false;
+   }
+ 
+   static isSelfHost(): boolean {
```
BlackDex, line 104: Should we make sure this also returns `false`?
Not sure where this is used and if we need to do that.

BlackDex (changes requested):

tessus:
Ok, so is this missing endpoint in the server for a new feature or will it break current functionality when installing the v2023.2.0 client?

What I am wondering is what will happen, if I install the v2023.2.0 clients and use them with the current version of vw server (or the main branch).

stefan0xC:
You can't change the collections of an entry, so the organization view will just look like this:
org_notloading

tessus:
Thanks, thus I won't update the clients. Hmm, but the other clients don't have that page anyway, so it's probably just the web client. I'll test on a VM to update the desktop client

BlackDex:
> You can't change the collections of an entry and the (new) organization view just looks like this: org_notloading [ org_notloading ](https://user-images.githubusercontent.com/509385/219207518-2a0322fd-5e01-4675-ab1b-b3f7b97af293.gif) [ ](https://user-images.githubusercontent.com/509385/219207518-2a0322fd-5e01-4675-ab1b-b3f7b97af293.gif)

It can be partially fixed by creating that endpoint and just redirect it to `get_org_collections(org_id, headers, conn).await`.

But there are a lot of other items changed, like the groups show which collections, the collections show which groups they belong to etc.. So, there are some more changes we need to make for it to work fully i think.

tessus:
But how is bw handling this? people who use the self-hosted system must also upgrade the server when they upgrade to 2023.2.0? Somehow this strategy is a bit flawed, because it is rather fragile. I'm also a bit fuzzy as to why they change the API this often. I've been using bw (clients) for less than a year and there were already 2 or 3 breaking changes.

BlackDex:
> But how is bw handling this? people who use the self-hosted system must also upgrade the server when they upgrade to 2023.2.0? Somehow this strategy is a bit flawed, because it is rather fragile. I'm also a bit fuzzy as to why they change the API this often. I've been using bw (clients) for less than a year and there were already 2 or 3 breaking changes.

They handle both front and backend. And they release/deploy the server/backend first and frontend later.

For self-hosted they have a script which Installs the correct versions together. That is actually the same as we do, only we need to build the vault first and check the PR's what has changed and what we need to add or update. Not only the clients repo, but also the server of course.

patches/v2023.2.0.patch
jjlin: Looks like `isSelfHost()` now calls the (recently-added) static version of itself, so the static version (or both versions) should be modified to return false. IIRC, if `isSelfHost()` returns true, then the web vault performs licensing checks in various places.

stefan0xC: Thanks for checking. I've set both to `false` for now.

PR #119, opened 2023-03-21
update web-vault to v2023.3.0

stefan0xC:
there was a new web-vault release [web-v2023.3.0](https://github.com/bitwarden/clients/releases/tag/web-v2023.3.0)

https://github.com/dani-garcia/vaultwarden/issues/3350 is still missing but aside from that I think the new interface should work without issues.

also the character limit for (new) passwords has been set to 12 characters.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,8 +28,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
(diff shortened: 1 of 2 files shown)


dani-garcia (approved):

tessus:
@stefan0xC may I ask if this web-vault works with the current main branch? except the missing bulk delete of course. But that feature was already in the previous web-vault release, correct? So nothing changes when people upgrade to this web-vault.

BlackDex:
From my quick tests there was nothing significant changed, and works on the current testing/main code.

PR #120, opened 2023-03-25
hide sso login link again

stefan0xC:
On the login page the current CSS rule `a[href$="/sso"]` for hiding the Enterprise single sign-on button does not match anymore because the email gets added to the end like so `#/sso?email=<email>`:
vaultwarden-login-sso-button

By targeting the `routerlink="/sso"` attribute instead the button should stay hidden again.


dani-garcia:
Good find, thanks!

PR #123, opened 2023-03-28
Moved images into directory structure

### LINKED ISSUE #122: Would you mind offer the resources to be in the correct tree structure?
Currently resources to patch in are in a flat folder (https://github.com/dani-garcia/bw_web_builds/tree/master/resources) and via script they are placed on the correct path (https://github.com/dani-garcia/bw_web_builds/blob/master/scripts/apply_patches.sh#L17). Would it be possible to make a tree directory structure for those resources?

This would allow other build systems to depend on bw_web_builds to recursively copy those files into the right places. And as well for yourself the copy script could just be a `cp -r ... ...`

BlackDex:
As requested in #122 moved all the image files into a correct directory structure. This does seem a bit logical indeed. And it makes it more clear where which replaced image goes. It also makes the `cp` command less error prone.

Fixes #122


stefan0xC (approved):
LGTM

PR #126, opened 2023-05-09
update web-vault to v2023.4.1

stefan0xC:
New web-vault release: [web-v2023.4.1](https://github.com/bitwarden/clients/releases/tag/web-v2023.4.1)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,8 +28,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2023.4.0
-ARG VAULT_VERSION=01e3a32d5098a50c110f599406e116794b0253c8
+# Using https://github.com/bitwarden/clients/releases/tag/web-v2023.4.1
+ARG VAULT_VERSION=4f4715fd0d7fc2e224a5efd5c6f8ce942922f067
 
 WORKDIR /vault
 RUN git init


tessus:
@BlackDex can you trigger a release? Would it make sense to create a GH action to create releases? e.g. it's even possible to sign packages via a gh action.

tessus:
Scratch that. 2023.4.2 just came out

BlackDex:
The .2 release didn't had any changes to the vault code or self. It was just a ci/cd change.

tessus:
ah, ok. Yep, I didn't check what changes there were in .2. I only saw `bug fixes`. They probably should have added infra or CI/CD in parenthesis... ;-)

BlackDex:
@tessus , released and i also signed the tar file using my gpg key. Not sure if that will causes issues for some people.
But my key is valid, and I also use it to sign all my git commits, so i think it should be ok.

tessus:
Awesome, thanks.
I don't think using your key should be an issue. People just need ro add your key, when they use a script to verify the .tar.gz. That's a one time thing.
It's not that people have to rewrite their pipeline in an effort of a few hours.... 😁

tessus:
@BlackDex the diagnostics page shows the following for the version: `2023.4.1^{}`

The `vw-version.json` is wrong. Not sure how that happened.

BlackDex:
That is strange indeed ill have a look

tessus:
yep, `git -c 'versionsort.suffix=-' ls-remote --tags --sort='v:refname' https://github.com/dani-garcia/bw_web_builds.git 'v*' | tail -n1 | grep -Eo '[^\/v]*$'` is giving this result.

Strange, we fixed that the last time. But I think that git has some changes recently. Hmm. I'll create a patch.

BlackDex:
I think that i created the tag a bit wrong maybe?

tessus:
Nope, the tag is ok. Argh, you replaced the sed with grep for compat, which makes this a bit more complicated. I can't use cut, because rev does not exist, and who knows if awk works the same on macos.

I am sorry, but making this compatible with macos was not a good idea. Just force people to install gsed if they want to build on macos.

BlackDex:
There are other ways. Ill check.

tessus:
I would just go with gnu sed which works on pretty much all sytems other than macOS.
I don't think it is too much to ask for people to install gnu sed on their macos. They have to install other tools as well. eg. automake, autoconf, ... or in vw case, they have to install rust.

stefan0xC:
> I think that i created the tag a bit wrong maybe?

I think you created an annotated tag and this shows when I run `git ls-remote --tags` there is second tag (the actual commit) because the annotated tag [is dereferenced](https://stackoverflow.com/questions/12938972/what-does-mean-in-git).

```
3dc2c6b932d3e8ec5ed63b54727b3eedd4ac1213	refs/tags/v2023.4.1
f01c4b28f62f184635e52ce06d27da78b297b45a	refs/tags/v2023.4.1^{}
```

So we just have to add `--refs` to the command and we should be fine.

tessus:
I just saw the same thing. `git tag` shows `v2023.4.1` but `ls-remote` shows the `^{}`. Good catch.

BlackDex:
@stefan0xC , and then, also update it to v2023.4.2 when you are at it :)

BlackDex:
@tessus, For a nice macOS working grep, this would be the sollution.

```bash
grep -Eoi '[^\/v]([0-9]+\.[0-9]+\.[0-9]+[a-z]|[0-9]+\.[0-9]+\.[0-9]+)'
```

tessus:
When you gpg sign a tag it is automatically annotated.

BlackDex:
> When you gpg sign a tag it is automatically annotated.

Looks like i didn't do that for both tags. Also, seems like that was never done before either. Ill see if i can use `-s` next time. All my commits are signed at least.

PR #128, opened 2023-06-01
Update web-vault to v2023.5.0

dpinse:
New web-vault release: [web-v2023.5.0](https://github.com/bitwarden/clients/releases/tag/web-v2023.5.0)


BlackDex (changes requested):
There are some patches missing, like the Name change from Bitwarden to Vaultwarden, the logo change.

Also, i see a new `region` select box on the login page, which doesn't seem to do anything, i think we might want to remove that from view also.

BlackDex (approved):

BlackDex:
Thanks!

PR #130, opened 2023-06-22
update web-vault to v2023.5.1

stefan0xC:
new web-vault release: [web-v2023.5.1](https://github.com/bitwarden/clients/releases/tag/web-v2023.5.1)


dani-garcia (approved):
Looks good, not sure if it's worth it to make a new release though, because the only changes are about the region selector which is hidden in our builds anyway

PR #131, opened 2023-06-22
add "web vault" to clarify version number in footer

stefan0xC:
Currently the web vault just displays the version number in the footer which seems to be confusing for many vaultwarden users as our server version number differs greatly from [Bitwardens versioning](https://bitwarden.com/help/versioning/).

So I was thinking that this might become a bit clearer if the footer says something like:

> Version 2023.5.1 (Web vault)

What do you think? I mean it would maybe be nicer if Bitwarden itself would display both web-vault version as well as the server version (by e.g. getting the version from `/api/config`) but that would probably be too big a change for our patch file.

### DIFF
--- a/patches/v2023.5.1.patch
+++ b/patches/v2023.5.1.patch
@@ -168,19 +168,22 @@ index 7b9a7b9477..1a26e31160 100644
  
    copyToClipboard(text: string, options?: any): void | boolean {
 diff --git a/apps/web/src/app/layouts/footer.component.html b/apps/web/src/app/layouts/footer.component.html
-index 8cacb4ceba..37d7cfb031 100644
+index 8cacb4ceba..eee2728221 100644
 --- a/apps/web/src/app/layouts/footer.component.html
 +++ b/apps/web/src/app/layouts/footer.component.html
-@@ -1,6 +1,6 @@
+@@ -1,8 +1,9 @@
  <div class="container footer text-muted">
    <div class="row">
 -    <div class="col">&copy; {{ year }} Bitwarden Inc.</div>
 +    <div class="col">Vaultwarden (unofficial Bitwarden&reg; server)</div>
      <div class="col text-center"></div>
      <div class="col text-right">
++      ({{ "webVault" | i18n }})
        {{ "versionNumber" | i18n : version }}
+     </div>
+   </div>
 diff --git a/apps/web/src/app/layouts/frontend-layout.component.html b/apps/web/src/app/layouts/frontend-layout.component.html
-index 84608acff0..9fdb254237 100644
+index 84608acff0..20f7a9666e 100644
 --- a/apps/web/src/app/layouts/frontend-layout.component.html
 +++ b/apps/web/src/app/layouts/frontend-layout.component.html
 @@ -1,6 +1,6 @@
@@ -191,12 +194,13 @@ index 84608acff0..9fdb254237 100644
      <bit-menu #environmentOptions>
        <a bitMenuItem href="https://vault.bitwarden.com" class="pr-4">
          <i
-@@ -29,6 +29,6 @@
+@@ -29,6 +29,7 @@
      <br />
    </div>
  
 -  &copy; {{ year }} Bitwarden Inc. <br />
 +  Vaultwarden (unofficial Bitwarden&reg; server)<br />
++  ({{ "webVault" | i18n }})
    {{ "versionNumber" | i18n : version }}
  </div>
 diff --git a/apps/web/src/app/layouts/navbar.component.html b/apps/web/src/app/layouts/navbar.component.html


tessus:
Yes, their clients all report the version a bit differently. The only client that actually showed the server version was the browser plugin, but it seems they removed it again.
I hoped they would add it this info to the other clients, but no, they removed it from the browser plugin. What a mess.

dionysius:
I believe it's good practice to hide versions in production, although it wouldn't matter here much as they are just web files. Having version output in the admin panel would be beneficial though.

tessus:
I believe in this case the version number of the web vault should be shown to the end user, as it currently is done. The vaultwarden API server version does not need to be shown on the web vault pages and that was never the idea.

I do not understand how it can be confusing to people as the OP mentioned. When they install a specific release of the web vault, why would they be confused? Then the version they installed is shown.
If someone installs the web vault, they most likey also installed vaultwarden, so they know the version of vaultwarden. Where is the confusion? Don't they understand that bitwaren and walwarden has 2 components. A server and clients (web, desktiop, mobile, cli). What is so complicated? It's not OpenStack where there are over 20 different components.

BlackDex:
The versions of Vaultwarden and Bitwarden are all fully accessable.
Either directly on the web-vault or by going to `/api/config`.

Both versions are already visible in the admin interface, so that is covered.


Regarding this change it self, i actually do not think it will help. People just do not read in these cases and quickly scan the page and see a version there. Also, Bitwarden it self is in sync with both server and web-vault version numbers most of the time, which also cause strange confusions something in regards to Vaultwarden.

stefan0xC:
The confusion arises because many users install a container without a version number (i.e. they'll pull `vaultwarden/server:latest`) and then look at the web-vault and then only see
Screenshot 2023-08-08 at 21-10-22 Vaultwarden Web Vault

How should they know that the version number refers to the web-vault and not Vaultwarden itself? Why should users check `/api/config`? (I mean they could also check the logs or run `vaultwarden --version` as explained in https://github.com/dani-garcia/vaultwarden/issues/3082 to get the correct vaultwarden version)

Side note: the web-vault version reported in the admin panel (which you have to enable) could apparently (cf. #134) be the wrong version as vaultwarden checks the file `vw-version.json` first (which is generated from the latest `bw_web_builds` release). I.e. this only works for our releases and not if I build an older (or newer) version of the web-vault. I.e.

I don't think that it will help insofar that we will never have people specifying the wrong version when they report an issue and I don't even think it's that big of an issue when they do but I think we could alleviate the confusion a bit by clarifying that the version refers to the web-vault.

BlackDex:
As a remark on the wrong version, that was a user error. We check both files, and the user build the vault him self, and used wrong logic into getting the version tag. That is not something we can help hehe.

My mention of the `/api/config` was in regards of hiding versions, noting that it is just available for everyone to see and hiding has no use.

It is a minor change indeed. But it's a change again hehe. It shouldn't be that big of an issue though.

stefan0xC:
@BlackDex I don't think it's an user error though. If you build `web-v2033.7.1` it will have a `vw-version.jon`  with `{"version":"2023.5.0"}` (whereas `version.json` will be correct). Not sure why we (still) need both files?

BlackDex:
> @BlackDex I don't think it's an user error though. If you build `web-v2033.7.1` it will have a `vw-version.jon` with `{"version":"2023.5.0"}` (whereas `version.json` will be correct). Not sure why we (still?) need both files?

The tag of the build is used, that is why it's still on the older version.
During the CI build it has the new tag and uses that.

The reason we do that is because sometimes, although this happened for a while now, we need to release a patched version which we tend to append with a letter. That will then be shown in the admin interface.

tessus:
Unfortunately I have learned that explaining the obvious or trying to make the obvious even more obvious is not going to work and won't change a thing.

There is also a reason why the issue template asks for the support string from the `/diagnostics` admin page.

This PR cannot cure ignorance nor stupidity.

BlackDex:
@stefan0xC
Ok, I do not think it will harm anyone or anything to have this more clear.
But... for me it sounds more logical to have the web-vault text in-front of the version number, instead of behind.

tessus:
> But... for me it sounds more logical to have the web-vault text in-front of the version number, instead of behind.

In this case the web vault text should not be in parenthesis.

stefan0xC:
I'd opted to keep it parenthetical `(Web vault) Version 2023.5.1` because without them it looks weird to me for some translations. I.e. in the english translation `Web vault Version 2023.5.1`. Having the translated string in parentheses is hopefully less obtrusive for most languages. Maybe it would be better to use `pageTitle` instead of `webVault`?

tessus:
IMO it shuld only say: `Web vault 2023.5.1`

There is no need for the word `Version`.

tessus:
I can only say that the current footer looks not very nice:

image

stefan0xC:
Not sure if we need to keep the `(unofficial Bitwarden® server)` but if not I think we could maybe switch it to:

> Vaultwarden Web Vault
> Version 2023.7.1

BlackDex:
That will also cause confusion if we release a patched version.
Also, we need to keep the Bitwarden text to indicate it's from Bitwarden.

tessus:
The word `Version` was useful previously, but now you have added the product name, thus the word `Version` is superfluous.

Having the product name in parenthesis at the begining of the sentence/line is a rather strange syntactic construct. Text in parenthesis refers to something mentioned previously in a sentence, not the other way around.

Please let me stress that it is not that important how the footer looks like, but currently it's just not very professional.

PR #133, opened 2023-07-11
feat: Bump web-vault to v2023.7.1

GeekCornerGH:
Requires:
- [ ] dani-garcia/vaultwarden#3592
- [x] dani-garcia/vaultwarden#3686


stefan0xC:
Um, [web-v2023.7.0](https://github.com/bitwarden/clients/releases/tag/web-v2023.7.0) has not even been released and I also think that it would most likely be based on the current `rc` branch not a specific commit in the `master` branch, wouldn't it?

Your patch file also has some issues: i.e. the name is wrong (it should be just `patches/v2023.7.0.patch`) and it cannot be applied (due to CLRF line endings and an error on line 25 where you have somehow replaced `©` with `┬®`). Also comparing it to the previous patch file it seems to be missing some important changes like setting `showTerms = false`, not removing the environment-selector or fixing the menu-component by adding `tw-overflow-x-auto`.

Dockerfile
```diff
@@ -28,8 +28,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2023.5.1
-ARG VAULT_VERSION=ee8e3f062cd4a9dd82d325940a54e8c0980c5b2f
+# Using https://github.com/bitwarden/clients/releases/tag/web-v2023.7.0
+ARG VAULT_VERSION=a1f6d19ab77c27cd5d61356a66ed5b60a28514cf
```
stefan0xC, line 32: Now that it has been released, this should be 
```suggestion
ARG VAULT_VERSION=d43aff0e5dd1796e95798ea03895f7957b9779ad
```

GeekCornerGH:
Thanks for the feedback. Not sure how the Copyright symbol got replaced, lines ending are probably related to a misconfiguration of github.dev tho. Will check this asap, thanks for the feedback

BlackDex:
I also see several other changes which have nothing to do with the normal patching we do.
They look like formatting changes, and only cause noise.

I also see that there are several changes made in a different way, like the following:
```diff
-    if (platformUtilsService.isSelfHost()) {
-      // Allow only valid email forwarders for self host
-      this.forwardOptions = this.forwardOptions.filter((forwarder) => forwarder.validForSelfHosted);
-    }
+    // Allow only valid email forwarders for self host
+    this.forwardOptions = this.forwardOptions.filter((forwarder) => forwarder.validForSelfHosted);
```

The reason we change `platformUtilsService.isSelfHost()` into `true`, is that the change is minimal, just one line and it probably stays the same even if the `this.forwardOptions` line changes, which currently could cause issues.

There is also a bit more removed on some places, one example:

```diff
-    const hasPremiumFromOrg = await this.stateService.getHasPremiumFromOrganization();
-    let billing = null;
-    if (!this.selfHosted) {
-      billing = await this.apiService.getUserBillingHistory();
-    }
-    this.hideSubscription =
-      !this.premium && hasPremiumFromOrg && (this.selfHosted || billing?.hasNoHistory);
+    this.hideSubscription = true;
```

The `const hasPremiumFromOrg` and `let billing` lines should still be there instead of being removed.

And, this patch also cloud break running on a sub-path in some situations i think.

```diff
-      const urls = process.env.URLS as Urls;
+      const urls = {base: getBaseUrl()};
       urls.base ??= this.win.location.origin;
       this.environmentService.setUrls(urls);
       this.environmentService.initialized = true;
```
While in theory it could be fine to leave the `url.base ??=` it might also break for some reason.
Removing that line will prevent issues i think (but not sure though).

And, it also seems to be missing some CSS patches to hide some fields. And there are also a few others which are allowed to be shown, like the externalId's for example.

GeekCornerGH:
Let me break down the comment.
> I also see several other changes which have nothing to do with the normal patching we do. They look like formatting changes, and only cause noise.

I guess that's just vs code autoformating issue. Thanks for the report (note for myself: [here](https://github.com/dani-garcia/bw_web_builds/pull/133/files#diff-7addab047413bb3a985c02d9ff89c50c0a3b747d38909e79bb2ed170575d851cR63), and [here](https://github.com/dani-garcia/bw_web_builds/pull/133/files#diff-7addab047413bb3a985c02d9ff89c50c0a3b747d38909e79bb2ed170575d851cR147), mention me if I forget something)

> I also see that there are several changes made in a different way, like the following:
> 
> ```diff
> -    if (platformUtilsService.isSelfHost()) {
> -      // Allow only valid email forwarders for self host
> -      this.forwardOptions = this.forwardOptions.filter((forwarder) => forwarder.validForSelfHosted);
> -    }
> +    // Allow only valid email forwarders for self host
> +    this.forwardOptions = this.forwardOptions.filter((forwarder) => forwarder.validForSelfHosted);
> ```
>
> The reason we change `platformUtilsService.isSelfHost()` into `true`, is that the change is minimal, just one line and it probably stays the same even if the `this.forwardOptions` line changes, which currently could cause issues.

`isSelfHost()` will always return false, per the [v2023.5.1 patch](https://github.com/dani-garcia/bw_web_builds/blob/master/patches/v2023.5.1.patch#L159-L167)

> There is also a bit more removed on some places, one example:
> 
> ```diff
> -    const hasPremiumFromOrg = await this.stateService.getHasPremiumFromOrganization();
> -    let billing = null;
> -    if (!this.selfHosted) {
> -      billing = await this.apiService.getUserBillingHistory();
> -    }
> -    this.hideSubscription =
> -      !this.premium && hasPremiumFromOrg && (this.selfHosted || billing?.hasNoHistory);
> +    this.hideSubscription = true;
> ```
> 
> The `const hasPremiumFromOrg` and `let billing` lines should still be there instead of being removed.

They would just add dead code in our case. They aren't returned, nor exported.

> And, this patch also cloud break running on a sub-path in some situations i think.
> 
> ```diff
> -      const urls = process.env.URLS as Urls;
> +      const urls = {base: getBaseUrl()};
>        urls.base ??= this.win.location.origin;
>        this.environmentService.setUrls(urls);
>        this.environmentService.initialized = true;
> ```
> 
> While in theory it could be fine to leave the `url.base ??=` it might also break for some reason. Removing that line will prevent issues i think (but not sure though).

I'm just using the same thing as [v2023.5.1 patch](https://github.com/dani-garcia/bw_web_builds/blob/master/patches/v2023.5.1.patch#L128-L130)

> And, it also seems to be missing some CSS patches to hide some fields. And there are also a few others which are allowed to be shown, like the externalId's for example.

Beside externalId (my bad, I forgot this was a thing in vaultwarden nowadays), can you tell me what's missing?

> Besides all that, thanks for trying :)

You're welcome 😄

> Edit: A small tip, try to run a diff on the previous patch file and the new one, that helps to see if there are missing patches, or if there is too much patched or not.

I actually patched using v2023.5.1 using --rej then fixed conflicts manually, and added some slight adjustements (hide region selector for example)

BlackDex:
I can be short on this, I think @stefan0xC made some bad changes then :p, and @dani-garcia didn't spotted them.
Please try to use the `v2023.5.0 then` ;)

stefan0xC:
> I'm just using the same thing as [v2023.5.1 patch](https://github.com/dani-garcia/bw_web_builds/blob/master/patches/v2023.5.1.patch#L128-L130)

https://github.com/dani-garcia/bw_web_builds/blob/d4637d01d4793919f24948b3c6a9173abef06295/patches/v2023.7.0.patch#L170-L172 is not the same as https://github.com/dani-garcia/bw_web_builds/blob/54477263b789f5e6e6b0715369012e68b4564df3/patches/v2023.5.1.patch#L128-L130

GeekCornerGH:
You're right, I apologies for the confusion. I somehow skipped this line.

GeekCornerGH:
Latest commit should fix all requested changes

BlackDex:
I still think there are a few items which could be better. The less patches you make the better for future patch compatibility.
The more lines you change the more it could deviate from a future version, and could result in more manual work to do.

A few items:

Here `const hasPremiumFromOrg` and `let billing` do not need to be removed.
While it probably does not effect the workings, it could effect future patches to be less compatible.
```diff
diff --git a/apps/web/src/app/settings/settings.component.ts b/apps/web/src/app/settings/settings.component.ts
index 60e81b17e..8078db5df 100644
--- a/apps/web/src/app/settings/settings.component.ts
+++ b/apps/web/src/app/settings/settings.component.ts
@@ -53,12 +53,6 @@ export class SettingsComponent implements OnInit, OnDestroy {
   async load() {
     this.premium = await this.stateService.getHasPremiumPersonally();
     this.hasFamilySponsorshipAvailable = await this.organizationService.canManageSponsorships();
-    const hasPremiumFromOrg = await this.stateService.getHasPremiumFromOrganization();
-    let billing = null;
-    if (!this.selfHosted) {
-      billing = await this.apiService.getUserBillingHistory();
-    }
-    this.hideSubscription =
-      !this.premium && hasPremiumFromOrg && (this.selfHosted || billing?.hasNoHistory);
+    this.hideSubscription = true;
   }
 }
```


I'm not sure if this patch is really needed, it doesn't actually add anything extra and could only cause more manual work in the future.
```diff
diff --git a/apps/web/src/app/auth/trial-initiation/trial-initiation.component.ts b/apps/web/src/app/auth/trial-initiation/trial-initiation.component.ts
index 57a807c9a..abf3693a6 100644
--- a/apps/web/src/app/auth/trial-initiation/trial-initiation.component.ts
+++ b/apps/web/src/app/auth/trial-initiation/trial-initiation.component.ts
@@ -122,11 +122,6 @@ export class TrialInitiationComponent implements OnInit, OnDestroy {

       this.referenceDataId = qParams.reference;

-      if (Object.values(ValidLayoutParams).includes(qParams.layout)) {
-        this.layout = qParams.layout;
-        this.accountCreateOnly = false;
-      }
-
       if (this.trialFlowOrgs.includes(qParams.org)) {
         this.org = qParams.org;
         this.orgLabel = this.titleCasePipe.transform(this.org);
```


Same here in regards to patching as minimal as we can. Setting the `if` statement to `true` will ensure that if Bitwarden updates the `this.forwardOptions` we do not accidentally overwrite it with something else which could break the vault.
```diff
diff --git a/apps/web/src/app/tools/generator.component.ts b/apps/web/src/app/tools/generator.component.ts
index 425f5cca2..232b2c855 100644
--- a/apps/web/src/app/tools/generator.component.ts
+++ b/apps/web/src/app/tools/generator.component.ts
@@ -37,10 +37,8 @@ export class GeneratorComponent extends BaseGeneratorComponent {
       route,
       window
     );
-    if (platformUtilsService.isSelfHost()) {
-      // Allow only valid email forwarders for self host
-      this.forwardOptions = this.forwardOptions.filter((forwarder) => forwarder.validForSelfHosted);
-    }
+    // Allow only valid email forwarders for self host
+    this.forwardOptions = this.forwardOptions.filter((forwarder) => forwarder.validForSelfHosted);
   }

   async history() {
```

And for the `apps/web/src/app/layouts/frontend-layout.component.html` file, I think we should remove the `<environment-selector [hasFlags]="false"></environment-selector>` line, that will prevent the US/EU/SelfHost selectbox from showing up.


And last, the CSS also needs some changes.
There are some layout changes on the Business Settings page which now show the Business name, which we do not support.
And, we now do support the identifier (Though that isn't there anymore i think).

So the lines:
```css
/* Hide Business Name and Identifier */
app-org-account form div.form-group:nth-child(3) { @extend %vw-hide; }
app-org-account form div.form-group:nth-child(4) { @extend %vw-hide; }
```

Can be removed/changed into:
```css
/* Hide Business Name */
app-org-account form div bit-form-field.tw-block:nth-child(3) { @extend %vw-hide; }
```

I think after those changes, it probably is going to be fine :).
Will probably need to have another walk through the interface to see if there are some other CSS patches needed, but i think this is enough.

GeekCornerGH:
Git somehow forgot to pick up the css change, I guess everything is done now

BlackDex:
Ok, i think this patch is fine btw, sorry for the delay.
But since v2023.7.1 has been released, maybe adjust it for that version :)

GeekCornerGH:
Did something changed on .1?

BlackDex:
Some small bugfixes. But since we didn't merged yet, best to use the latest available right now.

GeekCornerGH:
@BlackDex Finally bumped :)

GeekCornerGH:
Thanks for the catch, i forgot about this ;)

Dockerfile
stefan0xC, line 1: The `Dockerfile` has the wrong line endings.

patches/v2023.7.1.patch
```diff
@@ -0,0 +1,337 @@
...
+     this.hasFamilySponsorshipAvailable = await this.organizationService.canManageSponsorships();
+     const hasPremiumFromOrg = await this.stateService.getHasPremiumFromOrganization();
+     let billing = null;
+-    if (!this.selfHosted) {
+-      billing = await this.apiService.getUserBillingHistory();
+-    }
+-    this.hideSubscription =
+-      !this.premium && hasPremiumFromOrg && (this.selfHosted || billing?.hasNoHistory);
+-  }
++    this.hideSubscription = true;
```
stefan0xC, line 111: You have removed a `}` so this does not compile.

patches/v2023.7.1.patch
```diff
@@ -0,0 +1,337 @@
...
++          "otpauth://totp/Vaultwarden:" +
+           Utils.encodeRFC3986URIComponent(email) +
+           "?secret=" +
+           encodeURIComponent(this.key) +
+-          "&issuer=Bitwarden",
++          "&issuer=Vaultwarden",
+         size: 160,
+       });
+     }, 100);
+diff --git a/apps/web/src/app/settings/settings.component.ts b/apps/web/src/app/settings/settings.component.ts
```
stefan0xC, line 97: Compared to the `v2023.5.1.patch` you are missing many changes here.

stefan0xC (changes requested):

patches/v2023.7.1.patch
```diff
@@ -0,0 +1,337 @@
...
+diff --git a/apps/web/src/app/app.component.ts b/apps/web/src/app/app.component.ts
+index 9a22ea20b..b18a9accd 100644
+--- a/apps/web/src/app/app.component.ts
++++ b/apps/web/src/app/app.component.ts
+@@ -83,7 +83,7 @@ export class AppComponent implements OnDestroy, OnInit {
+     private keyConnectorService: KeyConnectorService,
+     private configService: ConfigServiceAbstraction,
+     private dialogService: DialogServiceAbstraction
+-  ) {}
++  ) { }
```
stefan0xC, line 64: Also this change is not necessary I think.

GeekCornerGH:
Thanks for the catch @stefan0xC, will check this out tomorrow, if I can :)

patches/v2023.7.1.patch
GeekCornerGH: Which ones exactly?

GeekCornerGH: Found what changes were missing, for some reason, vsc was hiding some rej files

patches/v2023.7.1.patch
```diff
@@ -0,0 +1,467 @@
...
+   }
+ 
+   isDev(): boolean {
+-    return process.env.NODE_ENV === "development";
++    return false;
+   }
+ 
+   isSelfHost(): boolean {
+@@ -141,7 +141,7 @@ export class WebPlatformUtilsService implements PlatformUtilsService {
+   }
```
BlackDex, line 174: I'm missing the patch here. It should contain `return false`, seems only the `static` version is changed here.

patches/v2023.7.1.patch
```diff
@@ -0,0 +1,467 @@
...
++  content: '' !important;
++  width: 32px !important;
++  height: 40px !important;
++  display: block !important;
++  background-image: url(../images/icon-white.png) !important;
++  background-repeat: no-repeat;
++  background-position-y: bottom;
++}
++/**** END Vaultwarden CHANGES ****/
+\ No newline at end of file
```
BlackDex, line 417: I don't think this should be here.

BlackDex (changes requested):
There are still some adjustments needed I think.

patches/v2023.7.1.patch
GeekCornerGH: I don't get your point, The "newline" stuff is added by Git

BlackDex:
@GeekCornerGH, i don't know what you did here, but now i see some patches which are multiple times in the file.
Like:
- `/apps/web/src/app/app.component.ts`
- `/apps/web/src/scss/styles.scss`
- `/apps/web/src/app/auth/settings/two-factor-authenticator.component.ts`

etc.. etc..

Seems to start at line 56 all the way to line 393

GeekCornerGH:
To be honnest, i don't know what's wrong with my pc, let me retry

GeekCornerGH:
Should be good now.

BlackDex:
@GeekCornerGH, It's a Monday patch i think haha.
I still see the `\ No newline at end of file` there ;)

GeekCornerGH:
@BlackDex I guess the `\ No newline at end of file` is just added by git itself, it isn't in the code (I updated recently my Git)

BlackDex (approved):

tessus:
Is it possible to make `Log in with device` available, iff push_notifications are enabled (`PUSH_ENABLED=true`)?

BlackDex:
> Is it possible to make `Log in with device` available, iff push_notifications are enabled (`PUSH_ENABLED=true`)?

If you mean, only show that button if that feature is enabled, then I'm afraid not. It will currently show that button always.

It will just not work. Also, you need to use the new websocket way if working for this to work at all btw.

tessus:
> If you mean, only show that button if that feature is enabled, then I'm afraid not. It will currently show that button always.

Too bad, because it will tell you that something was sent to your device, even though that is not true. Nothing can be sent to a device, if push notifications are not enabled.

I am not sure whether I should enable them. Is the overhead worth it? I was happy without the push notifications until now. When I added or updated in web/desktop/browser, the data was replicated almost instantly. That's all I needed, since it's not that hard to use the pull down gesture to sync on mobile.

Have you enabled push notifications?

> It will just not work. Also, you need to use the new websocket way if working for this to work at all btw.

This is ok. I am already using the new rocket websockets.

GeekCornerGH:
> Too bad, because it will tell you that something was sent to your device, even though that is not true. Nothing can be sent to a device, if push notifications are not enabled.

PUSH_ENABLED sets mobile push notifications, not websockets

tessus:
> PUSH_ENABLED sets mobile push notifications, not websockets

I know, but it is required for it to work on mobile - at least according to the PR:

_Logging in on mobile, and approving on mobile also works fine now over mobile push notifications._

PR #141, opened 2023-09-06
update web-vault to v2023.9.1

stefan0xC:
New web-vault release: [web-v2023.8.3](https://github.com/bitwarden/clients/releases/tag/web-v2023.8.3)

As far as I've compared the changes between [web-v2023.8.2...web-v2023.8.3](https://github.com/bitwarden/clients/compare/web-v2023.8.2...web-v2023.8.3) the main bug fix seems to be making the client compatible with older servers again (that don't send `hasMasterPassword` if you don't use Bitwardens key connector).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,8 +28,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2023.8.2
-ARG VAULT_VERSION=b403f2bcc79426abb9d8f02c391b7c8158876960
+# Using https://github.com/bitwarden/clients/releases/tag/web-v2023.9.1
+ARG VAULT_VERSION=e20fa3a5b23b33b19a269c67db8b4cd204345bbb
 
 WORKDIR /vault
 RUN git init
--- a/patches/v2023.9.1.patch
+++ b/patches/v2023.9.1.patch
@@ -1,0 +1,515 @@
+diff --git a/apps/web/src/app/admin-console/organizations/create/organization-information.component.html b/apps/web/src/app/admin-console/organizations/create/organization-information.component.html
+index 6029cfd833..04324b7d19 100644
+--- a/apps/web/src/app/admin-console/organizations/create/organization-information.component.html
++++ b/apps/web/src/app/admin-console/organizations/create/organization-information.component.html
+@@ -12,7 +12,7 @@
+       <input bitInput type="text" formControlName="name" />
+     </bit-form-field>
+     <bit-form-field class="tw-w-1/2">
+-      <bit-label>{{ "billingEmail" | i18n }}</bit-label>
++      <bit-label>{{ "email" | i18n }}</bit-label>
+       <input bitInput type="email" formControlName="billingEmail" />
+     </bit-form-field>
+     <bit-form-field class="tw-w-1/2" *ngIf="isProvider">
+diff --git a/apps/web/src/app/admin-console/organizations/settings/account.component.html b/apps/web/src/app/admin-console/organizations/settings/account.component.html
+index c7ac9910ac..e03c8fedcb 100644
+--- a/apps/web/src/app/admin-console/organizations/settings/account.component.html
++++ b/apps/web/src/app/admin-console/organizations/settings/account.component.html
+@@ -15,7 +15,7 @@
+         <input bitInput id="orgName" type="text" formControlName="orgName" />
+       </bit-form-field>
+       <bit-form-field>
+-        <bit-label>{{ "billingEmail" | i18n }}</bit-label>
++        <bit-label>{{ "email" | i18n }}</bit-label>
+         <input bitInput id="billingEmail" formControlName="billingEmail" type="email" />
+       </bit-form-field>
+       <bit-form-field>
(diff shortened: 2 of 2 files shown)


BlackDex:
If that is the only fix, then I'm not sure we should update this already.
Better wait for 2023.9.x

stefan0xC:
It's not the only bug fix (the other one would be to also [support onboarding Key Connector users with existing master passwords](https://github.com/bitwarden/clients/pull/6082), which I'm not sure how relevant is for us).

stefan0xC:
New update: [web-v2023.8.4](https://github.com/bitwarden/clients/releases/tag/web-v2023.8.4)

I don't think [this change](https://github.com/bitwarden/clients/commit/40bd6e8f969eab4f57e05ee93565f2218cd7c7f7) would be important for us either.

stefan0xC:
update to new web-vault release: [web-v2023.9.0](https://github.com/bitwarden/clients/releases/tag/web-v2023.9.0)

stefan0xC:
Also I was wondering since https://github.com/bitwarden/clients/issues/6309 was closed, if we should change the Authy recommendation ourselves?

stefan0xC:
I'll update my PR to 2023.9.1 tomorrow (when I have time to test the changes). If I do should I rename the patch file too?

BlackDex:
> I'll update my PR to 2023.9.1 tomorrow (when I have time to test the changes). If I do should I rename the patch file too? (The update doesn't require updating the patch file.)

I think that is wise. There are probably a lot of changes.
But it also needs to be tested good. Since they now force people to update the KDF. Do we need to do something on the server side too?

Same for the email generation annonaddy, we probably need to change that too.

stefan0xC:
@BlackDex Do you know how old an account would have to be to use a legacy encryption key? I.e. with what `bitwarden_rs` image I'd have to create it?

And yeah, you're right. The API call ((`https://app.addy.io/api/v1/aliases`) fails because of CSP.

BlackDex:
@stefan0xC The KDF iteration value was increased here: https://github.com/dani-garcia/vaultwarden/pull/3163 and released in https://github.com/dani-garcia/vaultwarden/releases/tag/1.28.0, so, not that old when we changed this on the server-side at least.

Looking at the code from Bitwarden it looks like they updated the default for new accounts in v2023.2.0 (https://github.com/bitwarden/clients/commit/2563fbd57d790257819efa0ac25c77686d4ab419) which would make it v1.28 also because we release with that web-vault version.

BlackDex:
@stefan0xC, i just did a quick test, and it seems to work just fine with Node18, so it's probably ok to update, also since Bitwarden stated it was an oversight.

For the KDF message we need to probably update the `/api/config` endpoint to provide the right key to show the message.
I think we should display this to increase the security `display-kdf-iteration-warning	true` https://vault.bitwarden.com/api/config

stefan0xC:
@BlackDex I've tested by creating a new user with the `bitwardenrs/server:1.10` image and after updating to the current Vaultwarden `main` branch and using the `v2023.9.1` web-vault still works and I don't get the form that should be shown according to this PR https://github.com/bitwarden/clients/pull/6195

I don't think that a low kdf iteration counts as having a legacy key. If you do have that you will get a warning once you have logged in but that has already been the case at least since `web-v2023.8.2`:
Screenshot 2023-10-01 at 04-04-23 Vaults Vaultwarden Web 
(btw: This warning get's displayed even without `display-kdf-iteration-warning: true`.)

After looking into it a bit I think that we can't have any users using legacy encryption (as defined [here](https://github.com/bitwarden/clients/blob/web-v2023.9.1/libs/common/src/auth/login-strategies/login.strategy.ts#L194-L195)) because [vaultwarden always](https://github.com/dani-garcia/vaultwarden/blob/5cd40c63ed230bee6dd54b79a9a7e768ad87978f/src/api/identity.rs#L129C17-L129C17) has returned a `key` (c.f. the [check here](https://github.com/bitwarden/clients/blob/web-v2023.9.1/libs/common/src/auth/login-strategies/password-login.strategy.ts#L170-L172)).

stefan0xC:
added https://github.com/dani-garcia/vaultwarden/pull/3950 for the anonaddy rename.

stefan0xC:
The 2fa recovery link has been hardcoded to https://vault.bitwarden.com/#/recover-2fa in https://github.com/bitwarden/clients/pull/5779

Should we change it back to the [previous help link](https://bitwarden.com/help/lost-two-step-device/) until bitwarden has fixed this for self-hosted installs?

BlackDex:
If that is a current bug then yes, since that will affect our users.

stefan0xC:
I've also reported it upstream: https://github.com/bitwarden/clients/issues/6487

BlackDex (approved):

reysonk:
@BlackDex when will the release be?

BlackDex:
Sorry, totally forgot this hehe :)

BlackDex:
Also, this will not mean it will also be included right away in Vaultwarden.

reysonk:
> Also, this will not mean it will also be included right away in Vaultwarden.

no problem, I will be able to build my build temporarily and wait for the official release.

BlackDex:
The web-vault has been released already. So you can use the one from the releases page

reysonk:
> The web-vault has been released already. So you can use the one from the releases page

I mean the VW assembly with the new version, created manually for myself so far :)
https://github.com/dani-garcia/vaultwarden/pull/3989

PR #145, opened 2023-10-31
update web-vault to v2023.10.0

stefan0xC:
New web-vault release: [web-v2023.10.0](https://github.com/bitwarden/clients/releases/tag/web-v2023.10.0)

Since they have updated the `/api/plans` endpoint again, I've <del>added</del> <ins>commented</ins> out some code to not call this endpoint at all and just always select the free plan.

I've also removed the new `product-switcher` component from the navbar since the secrets manager is not (yet) available for self-hosted (as far as I know) and is also not released under the terms of the GPLv3.

A further change would be to rename the `Bitwarden Send` to `Vaultwarden Send`. Since it would be to cumbersome to rename each occurrence of "Bitwarden" in the localized strings, we cannot really change the `sendCreatorIdentifier` message but I was wondering if we should remove the `sendAccessTaglineProductDesc` as well? (I've left it in for now but I don't think it's necessary).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,8 +28,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2023.9.1
-ARG VAULT_VERSION=e20fa3a5b23b33b19a269c67db8b4cd204345bbb
+# Using https://github.com/bitwarden/clients/releases/tag/web-v2023.10.0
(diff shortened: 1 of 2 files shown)


stefan0xC:
I've not yet looked into further changes (i.e. passkey support) but I think they would require us to change vaultwarden a bit as it's not a separate item type anymore but an additional field to the login item, last time I checked.

BlackDex:
> I've not yet looked into further changes (i.e. passkey support) but I think they would require us to change vaultwarden a bit as it's not a separate item type anymore but an additional field to the login item, last time I checked.

Yea, so we basically need to remove/revert those changes done before and make it work as it should work now.

BlackDex (approved):

PR #146, opened 2023-12-07
Updated latest patch file to work with the current upstream

Aldenar:
Current upstream commit: 77c54c2765fac5ca9040fbc279654eb47caaa2f8

However, some functionality definitely broke, so further testing and modifications are required

I found that at least creating an organization throws a "this.selectPlan is undefined" error and the org. doesn't get created.

Tested with NodeJS v20.5.1 and full host build.


BlackDex:
Closing this in favor of #147

PR #150, opened 2024-01-10
update web-vault to v2024.1.0

stefan0xC:
New web-vault release: [web-v2024.1.0](https://github.com/bitwarden/clients/releases/tag/web-v2024.1.0)

The creation of organizations has been refactored a bit but I think I made the necessary changes.

Flexible collections are still behind a feature flag and the new passwordless login with webauthn is too. 
To develop/test the latter, we can enable `passwordless-login` using the `EXPERIMENTAL_CLIENT_FEATURE_FLAGS` but I have not done so (we are still missing the required endpoints for this feature).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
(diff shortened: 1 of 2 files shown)


BlackDex (approved):

BlackDex:
@stefan0xC Thanks!
Did a quick test my self too, and it seems to work fine.

Now i was also able to test my new release make commands, and they worked like a charm!

PR #153, opened 2024-01-18
revert patch to fix max collection check

stefan0xC:
Apparently I made a mistake: https://github.com/dani-garcia/vaultwarden/discussions/4253

I've reverted the change and tested it (and it should just work). Sorry. :see_no_evil:

### DIFF
--- a/patches/v2024.1.0.patch
+++ b/patches/v2024.1.0.patch
@@ -420,7 +420,7 @@ index baff9d1b68..25f99cb40a 100644
      </div>
    </div>
 diff --git a/apps/web/src/app/vault/individual-vault/vault.component.ts b/apps/web/src/app/vault/individual-vault/vault.component.ts
-index 964debb536..12b23f8888 100644
+index 5ca857dae6..bc7568f5f4 100644
 --- a/apps/web/src/app/vault/individual-vault/vault.component.ts
 +++ b/apps/web/src/app/vault/individual-vault/vault.component.ts
 @@ -188,11 +188,7 @@ export class VaultComponent implements OnInit, OnDestroy {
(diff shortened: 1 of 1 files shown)


BlackDex (approved):

BlackDex:
Cool, thanks.

Now I'm also able to test a letter suffix to the versions haha ;)

tessus:
@BlackDex hmm, wouldn't that be `a` first?

BlackDex:
We never used `a` always a `b` as postfix if the version of the vault didn't changed.

tessus:
I see. I just had the old versioning scheme of openssl in my head and the started with an `a`.

BlackDex:
They do start with a, but they never started without a letter ;).
Hinge why we now use a `b` since its the second release of v2024.1.1

tessus:
My thought process was (in case you are wondering) that a suffix, like a letter, for versioning was used if it didn't fit the semantic versioning (or could also be used for a different reason) and denoted that it was different from the base version. Thus `1.0.0a` is different than `1.0.0`. Starting with `a` would have been logical to me, because it is the first letter in the alphabet. But in reality it doesn't matter where one starts. It's just that it looks a bit like `1.0.0` and `1.0.0a` could be the same, which is weird. It feels like one release was skipped. Just my personal opinion, nothing more.

But it also makes sense to say `b` is the 2nd release, because `b` is the second letter. Well, although in IT 0-based indexing is often used. Anyway, all good and thanks for the info.

BlackDex:
So what is the 0 based letter then ;) haha.

To be fair, i was doubting between `a` and `b`, and i just selected what previously was used.
But it's good to discuss this.

tessus:
> So what is the 0 based letter then ;) haha.

Haha. No I meant it like this:

1.0.0        Base version - index 0 (1st release of that version)
1.0.0a       index 1 (2nd release of that version)
1.0.0b       index 2 (3rd release of that version)

> But it's good to discuss this.

Yep, it's always fun - at least for me.

PR #155, opened 2024-02-04
revert patch to edit org info

stefan0xC:
Sorry, another issue introduced by me. 🫣

fixes [this external bug report](https://vaultwarden.discourse.group/t/cant-change-organization-name-and-emil/3318?u=stefan0xc)

### DIFF
--- a/patches/v2024.1.2.patch
+++ b/patches/v2024.1.2.patch
@@ -25,18 +25,17 @@ index 83e0e35cf3..9385c4071c 100644
        </bit-form-field>
        <bit-form-field>
 diff --git a/apps/web/src/app/admin-console/organizations/settings/account.component.ts b/apps/web/src/app/admin-console/organizations/settings/account.component.ts
-index 10aa1224e2..d3219be672 100644
+index 10aa1224e2..3b6f3a702a 100644
 --- a/apps/web/src/app/admin-console/organizations/settings/account.component.ts
 +++ b/apps/web/src/app/admin-console/organizations/settings/account.component.ts
-@@ -183,7 +183,7 @@ export class AccountComponent {
+@@ -182,6 +182,7 @@ export class AccountComponent {
+   };
  
    submitCollectionManagement = async () => {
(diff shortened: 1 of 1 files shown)


BlackDex:
Luckily we didn't released 1.30.4 yet 😂

stefan0xC:
I think I'll have to get more familiar with the web-vault because the whole `isSelfHosted` stuff is very confusing to me. :sweat_smile:

BlackDex:
That is more because of the whole organization licensing part.
The non-self-hosted allows to create orgs without a license file, the self-hosted need a file.

So, we mimic a bit more the SaaS version of the web-vault this way.

stefan0xC:
Yeah, but I think it would be less confusing (at least to me) if it would default to `true` and we only made the necessary changes wherever we had to instead of pretending to be the cloud version and having to decide where it's necessary to revert the check for self hosted. :thinking:

BlackDex:
Worth looking into i think.

patches/v2024.1.2.patch
BlackDex, line 1: On line 35 and 36 you also did some inverting of the selfHosted value.
Maybe it's better to leave that as is, and put `return;` above the `if` statement?
And maybe add a comment there like `return; // flexible collections are not supported by Vaultwarden`

That way it's clear we do not support that part, and it will not break if somehow that values changes hehe ;).

BlackDex (changes requested):

stefan0xC:
I'll do a separate PR once this is merged for rewriting the patch file. Unless we are fine with changing the background color of the navbar? (A self hosted Bitwarden instance uses [a different background color](https://github.com/bitwarden/clients/blob/c97e5b3d1ffe201fe807896a22e4f731fba73d65/apps/web/src/app/layouts/navbar.component.html#L1) (`#1a3b66`) from the default `#175ddc` which I would change to the color we use in the /admin navbar, i.e. `#212529`)

BlackDex (approved):

PR #157, opened 2024-03-06
update web-vault to v2024.3.0 (new vertical layout)

stefan0xC:
There have been two new web-vault releases yesterday: [`web-v2024.2.3`](https://github.com/bitwarden/clients/releases/tag/web-v2024.2.3) and [`web-v2024.2.4`](https://github.com/bitwarden/clients/releases/tag/web-v2024.2.4) which would change the layout quite drastically (cf. https://github.com/bitwarden/clients/commit/38d8fbdb5a7498b77d445bded30be00b9661a054).
[UPDATE] The article that explains the layout change has been published: https://bitwarden.com/blog/bitwarden-design-updating-the-navigation-in-the-web-app

The patch file differs further from previous files because with the layout change there are more drastic changes. I.e. the navbar at the top was removed (and replaced with a vertical bar), there's  no footer once you are logged in and the Organizations link at the top to manage organizations was moved to the product-switcher.  Since the svgs will be embedded into the page I've created two logos to differentiate between the `Password Manager` and the `Admin Console` (which is currently only for managing organizations and should not be confused with the `/admin` panel.)

Password Manager:
vaultwarden-password-manager-logo 

Admin Console:
vaultwarden-admin-console-logo

There are also some further changes that we should add to keep vaultwarden compatible with the web-vault.
E.g. flexible collections (https://github.com/bitwarden/clients/commit/0c3b569d0e23dbdced735d3447653fa9b18b463a), login with passkey support (https://github.com/bitwarden/clients/commit/3a0603a837fc20c10a124983f926766293ee63ee), 

Since the feature flag for login with passkey has been removed (i.e. the option will be shown even though we don't have the necessary endpoints) I've hidden/removed the components for now.

The patch currently looks like this:

Light mode:
Bildschirmfoto am 2024-03-09 um 08 27 16
Bildschirmfoto am 2024-03-09 um 08 32 45

Dark mode:
Bildschirmfoto am 2024-03-06 um 05 31 52
Bildschirmfoto am 2024-03-06 um 05 30 34

As far as I've quickly tested it at least the core functionality (i.e. creating, modifying, deleting accounts, items, organizations, collections, sends, ...) still works.


stefan0xC:
I've added the two logos to the resources folder and created a basic script that replaces both logos (this currently requires a sed with in-place-editing capabilities).

Also changed the color scheme of the light mode a bit.

tessus:
@stefan0xC how complex is it to add the endpoints for logging into vw via passkeys? Is this something I can help you with? I doubt I can work on this alone, but if you start with it, I can work on smaller tasks you assign to me. would that help, or is this more work than someone doing this alone?

stefan0xC:
@tessus I'm not sure because I have not looked into it. If I were to tackle it I would also like to update the [`webauthn-rs`](https://crates.io/crates/webauthn-rs) crate to the latest version (which probably is a more complex task) but I have not had time to dive into it. This update is probably not required to implement the endpoints though so don't let it deter you if you want to take a crack at it.

BlackDex:
I'm not 100% sure, but the main reason we want to update to the latest `webauthn-rs` crate is also because it has better support for edge cases. The best would be if we could use the _Safe_ API version of it instead of use the _Core_ part manually.

Only, not sure if that can be done.

BlackDex:
I have some items which we need to address on the server side before we can release this:
 - `fn post_rotatekey` This doesn't work anymore (might be in previous version also btw).
    https://github.com/bitwarden/clients/commit/a62f8cd6522eec49d84d0f7c18cf315a8603edac
    
 - Do we need to exclude the duo-redirect connectors also from adding some security headers?
    Duo seems to work for me using the web-vault though. I'm not getting a redirect.
    And I think it is only for Organization Duo MFA, which we do not support.
    https://github.com/bitwarden/clients/commit/c481d929488984e74c73ca94194eae874b07fe04
    https://github.com/bitwarden/clients/commit/c91ceb20143a7a866c0e9bc2aeeca40335193f0b
    
 - This caught my eye also, not fully sure, but it might be linked only to the new flex collections?
   https://github.com/bitwarden/clients/commit/5c6245aaaef6453d3e6ee2e6dff9b5d805baafb9

 - Didn't we added support for vault-item-encryption already? They seem to have bumped the minimum server version.
   If that is the case, we need to update ours as well.
   https://github.com/bitwarden/clients/commit/c199f02d442504d2e78b582f3c045470cdc2b0c5

I'm not seeing any other issues here.
The rest seems to work ok.

BlackDex:
So, I'm fine btw with approving this PR, but I'm probably going to wait to release it until after we at least made adjustments to the key rotation code.

BlackDex (approved):

BlackDex:
Ow, they just release v2024.3.0 🤣

BlackDex (approved):

BlackDex:
Since the key-rotation is already broken in the other versions also, lets just merge and release :).
Thanks @stefan0xC !

tessus:
Is there a certain release required for this to work? e.g. `1.30.5` or a certain commit? I know that WIP PR would have been nice, but as you said it was already broken for a few releases.

BlackDex:
It works with the current testing release as far as i could tell.

tessus:
Ah, I see. Thanks for the info.

I am asking because I also create RPM packages for Fedora, but I guess they are mostly for myself (I don't track who downloads them). However, I usually have a `Requires: vaultwarden >= xxxx` in the web-vault package.

tessus:
@stefan0xC why do the password manager and the admin console use a different background color in the light theme, but use the same background color in dark theme?
Is this on purpose or an oversight?

stefan0xC:
@tessus the main background-color [is the same](https://github.com/bitwarden/clients/blob/da14d01062d2987e4599aa4c877f74ab7e7ddb89/libs/components/src/tw-theme.css#L5C1-L6C9)?

If you mean the side-panel: That's presumably a design decision by Bitwarden.

In light mode they use [--color-background-alt3](https://github.com/bitwarden/clients/blob/da14d01062d2987e4599aa4c877f74ab7e7ddb89/libs/components/src/tw-theme.css#L9) for the password manager (`rgb(18, 82, 163)`) which I have changed to `rgb(33, 37, 41)` and [--color-secondary-100](https://github.com/bitwarden/clients/blob/da14d01062d2987e4599aa4c877f74ab7e7ddb89/libs/components/src/tw-theme.css#L16) for the admin console (`rgb(240, 240, 240)`), which I have not changed.
https://github.com/dani-garcia/bw_web_builds/blob/ac7f24e355f537877e355b36ea41e744ad658aeb/patches/v2024.3.0.patch#L698-L719

tessus:
@stefan0xC I was talking about the side panel:

image

vs

image

But in dark mode they look the same. It's all good if this is not a bug. It just seemed off to me that the backgrounds are different in the light theme but the same in the dark theme.

misilot:
Does this PR bring in the Duo Universal prompt support? That came with web-vault 2024.2.3?

https://bitwarden.com/help/releasenotes/#2024-2-3

BlackDex:
> Does this PR bring in the Duo Universal prompt support? That came with web-vault 2024.2.3?
> 
> https://bitwarden.com/help/releasenotes/#2024-2-3

Technically the client side adds this feature, but for this to fully work we need to make changes to the server side too.
Which have not been done.

Gigaion:
Looks like Bitwarden came out with a v2024.3.1 with a nice feature add of bullk collection moving.

https://github.com/bitwarden/clients/releases/tag/web-v2024.3.1

stefan0xC:
@Gigaion I've already made a PR https://github.com/dani-garcia/bw_web_builds/pull/159

BlackDex:
Though i didn't find that feature out-off-the-box. So it probably needs a feature flag and server-side adjustments.

lukasj98:
I think the light theme for the panel would be better for the regular password manager and for the admin the dark could you change this

BlackDex:
You mean, light theme when on the personal manager, and the dark theme when on the organization manager?

That would be strange, and also not wanted in most cases i think, I have dark themes for almost everything for example.

lukasj98:
Why not the original color ?

lukasj98:
But I think it would be better if it would change according to the color scheme settings

lukasj98:
And dark mode doesn't work with new version

PR #159, opened 2024-04-04
update web-vault to v2024.3.1

stefan0xC:
New web-vault version: [web-v2024.3.1](https://github.com/bitwarden/clients/releases/tag/web-v2024.3.1)

I've not checked if we need to change anything else server-side. (Basic stuff still works, also running Vaultwarden in a alternate base dir, but have not tested everything, e.g. duo connector...)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2024.3.0
-ARG VAULT_VERSION=dbf0f1e328e73bbdea9f120c020770c9e117ad6d
+# Using https://github.com/bitwarden/clients/releases/tag/web-v2024.3.1
+ARG VAULT_VERSION=60e47aff88024ba9aad1a4ed7b63690b968bf7fc
 
 WORKDIR /vault
 RUN git -c init.defaultBranch=main init && \
--- a/patches/v2024.3.1.patch
+++ b/patches/v2024.3.1.patch
@@ -1,0 +1,797 @@
+diff --git a/apps/web/src/app/admin-console/organizations/create/organization-information.component.html b/apps/web/src/app/admin-console/organizations/create/organization-information.component.html
+index 6029cfd833..04324b7d19 100644
+--- a/apps/web/src/app/admin-console/organizations/create/organization-information.component.html
++++ b/apps/web/src/app/admin-console/organizations/create/organization-information.component.html
+@@ -12,7 +12,7 @@
+       <input bitInput type="text" formControlName="name" />
+     </bit-form-field>
+     <bit-form-field class="tw-w-1/2">
+-      <bit-label>{{ "billingEmail" | i18n }}</bit-label>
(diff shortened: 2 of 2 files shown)


BlackDex:
Looks like there are some changes in the event collections, but not sure that has an impact. 

There now is a bulk collection update option. 
https://github.com/bitwarden/clients/commit/bac0874dc0331661a0af194d7937a72509736483

That is probably something we might need to update our add something for on the server side.

stefan0xC:
> I just noticed that there is a call done for every org during login. `https://127.0.0.1/api/organizations/<org_uuid>/billing-status`, can we prevent this maybe?

Does this happen with `v2024.3.0` because I thought I've patched it out for this version.
Compare https://github.com/stefan0xC/bitwarden_clients/commit/c4de547d70ba7ab0e4be8bf933f88d24dc9ce410 vs https://github.com/stefan0xC/bitwarden_clients/commit/d6b1c8120d114dc34dfd0bd75d8796d5326d8074

stefan0xC:
Ah yes, just checked and it happens with `v2024.3.0`. I can backport the changes to the `v2024.3.0.patch`.

BlackDex:
Nea, i think this release can be the final one and added to Vaultwarden instead of the 2024.3.0.

BlackDex:
Since the new bulk collection feature is behind a feature-flag, lets merge this :).

tessus:
The more important question is whether a new API endpoint is required for the feature flag to work.

BlackDex:
> The more important question is whether a new API endpoint is required for the feature flag to work.

Probably yes, or at least changes to existing ones, since else a feature flag would not be needed.

Gerardv514:
In dark mode, under preferences screen you can’t see the field drop down indicators on each field, but they are viewable on light theme.

stefan0xC:
@Gerardv514 I don't think I have that problem (at least in Firefox):
Screenshot 2024-04-08 at 07-28-29 Preferences Vaultwarden Web
Screenshot 2024-04-08 at 07-30-14 Preferences Vaultwarden Web

Did you fully reload the page? I mean if you can recreate this with Bitwarden Cloud (https://vault.bitwarden.com/), I would suggest you create an issue in https://github.com/bitwarden/clients/issues with more details (i.e. OS/Browser version, etc) since this should probably be addressed upstream then.

Gerardv514:
👍 @stefan0xC Seems like this behavior only occurs on mobile browsers (tested on iPhone, Chrome & Safari browsers). Issue does not exist on computer browser. Minor issue, and probably exists upstream.

BlackDex:
> 👍 @stefan0xC Seems like this behavior only occurs on mobile browsers (tested on iPhone, Chrome & Safari browsers). Issue does not exist on computer browser. Minor issue, and probably exists upstream.

I'm not having issues on my mobile either.

Gerardv514:
Hmm odd..

IMG_3417
IMG_3416

BlackDex:
Screenshot_20240410-182611~2

Gerardv514:
@BlackDex what browser and device is your screenshot above from? I noticed that this is probably a desktop browser. Desktop browsers only show the down arrow, whereas the mobile browsers show an up and down arrow.

Gerardv514:
@stefan0xC question, when in Light Theme, why is the password manager side bar black (as if in dark theme) rather than the default blue?


image

PR #160, opened 2024-04-17
update web-vault to v2024.5.0

stefan0xC:
New web-vault release: [web-v2024.4.1](https://github.com/bitwarden/clients/releases/tag/web-v2024.4.1)

> [!NOTE]
I've skipped `v2024.4.0` because it only removed  the banner informing users about the layout change. The banner was replaced by one that shows a message if there are unassigned messages (https://github.com/bitwarden/clients/pull/8655). But for that to work we'd need to provide the `/api/ciphers/has-unassigned-ciphers` endpoint (https://github.com/bitwarden/server/pull/3967) in vaultwarden. (This feature is currently behind the `UnassignedItemsBanner` flag.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2024.3.1
-ARG VAULT_VERSION=60e47aff88024ba9aad1a4ed7b63690b968bf7fc
+# Using https://github.com/bitwarden/clients/releases/tag/web-v2024.4.1
+ARG VAULT_VERSION=ceffa2064213ce2ffbcbe69853e6285391bfffa0
 
 WORKDIR /vault
 RUN git -c init.defaultBranch=main init && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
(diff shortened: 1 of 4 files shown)


BlackDex:
Ah nice, lets test it out.
We also might want to update the meta features request on the other repo to maybe add some of these feature flags and maybe create discussions in the Idea section for these kind of features, same as for flexible collections.

That way we have somewhere to track missing features and also can note if someone starts working on it.

BlackDex:
@stefan0xC i totally forgot this one, sorry.
Maybe best to update to v2024.4.2 then?

BlackDex:
Seems there is a DUO issues with that version.
Not sure if that is also the case for this version, but it will hinder my tests to migrate to the new DUO way of login.

stefan0xC:
I have not really looked at this release yet. Just noticed an issue with getting billing metadata (which we have not implemented), so I've set it to `false` here:
https://github.com/dani-garcia/bw_web_builds/blob/916c10825fa9d2ebd073b2e823c5fc10d122a421/patches/v2024.4.2.patch#L33-L40

BlackDex:
Uch, on v2024.5.0 now haha.
I really need to find some time to do some testing and release this

stefan0xC:
I've updated the web-vault to the newest release. Regarding testing I think that we might want to add `playwright` tests to this repo but we should maybe discuss it in this thread: https://github.com/dani-garcia/vaultwarden/pull/4369#issuecomment-2110963900?

BlackDex:
> I've updated the web-vault to the newest release. Regarding testing I think that we might want to add `playwright` tests to this repo but we should maybe discuss it in this thread: [dani-garcia/vaultwarden#4369 (comment)](https://github.com/dani-garcia/vaultwarden/pull/4369#issuecomment-2110963900)?

Yea, that would be useful indeed.

BlackDex:
Thanks @stefan0xC, I implemented this web-vault in my latest PR to update the crates and Rust :).

PR #163, opened 2024-05-21
remove non-OSI license code to ensure it's not used in builds

### LINKED ISSUE #162: Is GPL-3.0 correct?
Hi,

I'm a Fedora/EPEL packager and was working towards getting Vaultwarden and the web side packaged up into official packages and all is/was going smoothly until it came to my attention that code from https://github.com/bitwarden/clients/tree/main/bitwarden_license is indeed used in bw_web_builds.

I tested by simply removing the directory as part of the build scripts, and unsurprisingly, the build fails.

I believe this renders GPL-3.0 incorrect, at least for part of the content, anything built upon the bitwarden-licensed code, and unfortunately bars the web GUI from inclusion into Fedora/EPEL and any other distros which follow strict OSI-approved license requirements.

The restrictions from https://github.com/bitwarden/clients/blob/browser-v2024.5.0/LICENSE_BITWARDEN.txt#L27-L31 specifically make the bitwarden license incompatible with OSI standards.

I'm not very familiar with node/TS so I don't know exactly how/what these files are doing so I apologize if I'm totally off base here.

jonathanspw:
fixes #162

### DIFF
--- a/patches/v2024.5.0.patch
+++ b/patches/v2024.5.0.patch
@@ -832,3 +832,15 @@ index 00ab2ff717..0950b9d787 100644
  
    --color-secondary-100: 240 240 240;
    --color-secondary-300: 206 212 220;
+diff --git a/apps/web/tsconfig.json b/apps/web/tsconfig.json
+index 543d7f25..5579906e 100644
+--- a/apps/web/tsconfig.json
++++ b/apps/web/tsconfig.json
+@@ -30,7 +30,6 @@
+   "files": [
+     "src/polyfills.ts",
+     "src/main.ts",
+-    "../../bitwarden_license/bit-web/src/main.ts",
+     "src/theme.ts"
+   ],
+   "include": [
--- a/scripts/checkout_web_vault.sh
+++ b/scripts/checkout_web_vault.sh
@@ -53,4 +53,7 @@ pushd "${VAULT_FOLDER}"
     git fetch --tags --depth 1 origin "${VAULT_VERSION}"
     # Checkout the branch we want
     git -c advice.detachedHead=false checkout FETCH_HEAD
+    # Make sure we don't use non-OSI code
+    # requires patching in next step
+    rm -rf bitwarden_license
 popd


scripts/checkout_web_vault.sh
```diff
@@ -53,4 +53,7 @@ pushd "${VAULT_FOLDER}"
     git fetch --tags --depth 1 origin "${VAULT_VERSION}"
     # Checkout the branch we want
     git -c advice.detachedHead=false checkout FETCH_HEAD
+    # Make sure we don't use non-OSI code
+    # requires patching in next step
+    rm -rf bitwarden_license
```
stefan0xC, line 58: This should probably be done in the `apply_patches.sh` script so it will also be removed from the Docker build.

Also if you remove this directory you should also exclude it from the generated patch file:
https://github.com/dani-garcia/bw_web_builds/blob/cc03d26ac0f47481067f6ae313375e198a7b1164/scripts/generate_patch_file.sh#L40-L41

patches/v2024.5.0.patch
```diff
@@ -832,3 +832,15 @@ index 00ab2ff717..0950b9d787 100644
...
+--- a/apps/web/tsconfig.json
++++ b/apps/web/tsconfig.json
+@@ -30,7 +30,6 @@
+   "files": [
+     "src/polyfills.ts",
+     "src/main.ts",
+-    "../../bitwarden_license/bit-web/src/main.ts",
+     "src/theme.ts"
+   ],
+   "include": [
```
stefan0xC, line 846: If I try to apply this patch file it will complain because of the line ending. If  you generate the patch file with `make generate-patch` this should be corrected.

scripts/checkout_web_vault.sh
BlackDex, line 58: This doesn't seem useful, that directory is not packed or used, only pulled as part of the repo. Removing this doesn't serve anything useful.

patches/v2024.5.0.patch
```diff
@@ -832,3 +832,15 @@ index 00ab2ff717..0950b9d787 100644
...
    --color-secondary-300: 206 212 220;
+diff --git a/apps/web/tsconfig.json b/apps/web/tsconfig.json
+index 543d7f25..5579906e 100644
+--- a/apps/web/tsconfig.json
++++ b/apps/web/tsconfig.json
+@@ -30,7 +30,6 @@
+   "files": [
+     "src/polyfills.ts",
+     "src/main.ts",
+-    "../../bitwarden_license/bit-web/src/main.ts",
```
BlackDex, line 843: This isn't needed anymore as this is already changed upstream https://github.com/bitwarden/clients/pull/8818/files

So, I also do not see the huge benefit here either. 
We already build the OSS version, which is determined by Bitwarden to be GPLv3 compatibel ans not using there Bitwarden Licensed code.

BlackDex (commented):
Thx for the PR, but i think it's not very useful to implement this because of the reasons given per change

stefan0xC (changes requested):
Personally I would remove the `bitwarden_license` references from all files for good measure but it should build just fine with only removing it from the `apps/web/tsconfig.json`.

Not sure if we really want to remove the `bitwarden_license` directory though as it would require additional handling in the scripts.

scripts/checkout_web_vault.sh
BlackDex: I think it's too much for something not used at all.

BlackDex:
> Personally I would remove the `bitwarden_license` references from all files for good measure but it should build just fine with only removing it from the `apps/web/tsconfig.json`.
> 

Which is already done upstream, so i do not see a use case for this until a new released version of the web-vault, at which point it will be solved automatically.

> Not sure if we really want to remove the `bitwarden_license` directory though as it would require additional handling in the scripts.

It will only complicate stuff with the current way of patching i think.

patches/v2024.5.0.patch
stefan0xC: I think the benefit would be the assurance that the compiled code is actually `GPLv3` compatible and not relying on the content in the `bitwarden_license` directory at all to be built, which might not always be the case. (E.g. in [`web-v2023.8.0`](https://github.com/bitwarden/clients/blob/web-v2023.8.0/apps/web/src/app/billing/organizations/secrets-manager/sm-subscribe.component.ts#L10) to [`web-v2023.9.2`](https://github.com/bitwarden/clients/blob/web-v2023.9.2/apps/web/src/app/billing/organizations/secrets-manager/sm-subscribe.component.ts#L10) the secrets manager logo was technically included as part of the compilation when it was located in the `bitwarden_license` directory) 

It also enables third party vendors to be able to ship the source code without bundling the non-free code into their packages as explained in #162. (If we were to create a soft fork, we could also easily remove the non-free code from our repository to be on the safe side.)

jonathanspw: Precisely.  For distros and probably others all assurances should be taken that no non-OSI code ends up in the distro by mistake.  That said, it's standard practice to "rm -rf" anything you want to make sure isn't included or built against in RPM spec files.

My intent is to use the published/built sources for web vault since it's not binary, which means I can't rm -rf at build time with a Fedora-specific patch.

Since the patch does no harm and removes what's not being used anyway, it addresses the concern by having multiple things that would have to fail for the code to get into the build by mistake, while having 0 impact on existing functionality/expectations.

BlackDex: I'm for the soft fork option but not currently in this repo. Though i still wonder if it isn't more work with upstream rebases etc. (But that is something i haven't looked at actually)

In general I agree that if possible without too much hassle to try and keep clear of any possible GPLv3 invalid build.

But seeing that upstream already made some changes regarding this i rather wait for a new release with those changes incorporated.

BlackDex:
Btw, if it makes releasing your package really more useful in the short term. Then i do not have any strong feelings against it. But if not, then i would rather wait for a new release to at least add the directory removal.

BlackDex:
Closing as fixed via #169

PR #164, opened 2024-05-25
Prepare for repo to org move

BlackDex:
This PR prepares this repo to be moved to the Vaultwarden organization.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,7 @@ WORKDIR /vault/apps/web
 RUN npm run dist:oss:selfhost
 
 RUN printf '{"version":"%s"}' \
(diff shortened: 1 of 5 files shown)


stefan0xC:
It's not a big deal because the patch file can still be applied but your commit seems to remove all trailing spaces from the patch set leading to unrelated changes in the patch file.

BlackDex:
> It's not a big deal because the patch file can still be applied but your commit seems to remove all trailing spaces from the patch set leading to unrelated changes in the patch file.

Ugh that is because my editor trims those by default.

BlackDex:
I removed the changes from the patch file since those might get updated in before this gets merged :)

toranih:
- - - [ ] # @stefan0xC

toranih:
- - - - @@sciyoshi @sciyoshi

PR #166, opened 2024-06-03
update web-vault to v2024.5.1

stefan0xC:
The new web-vault version [web-v2024.5.1](https://github.com/bitwarden/clients/releases/tag/web-v2024.5.1) just includes a small bug fix: https://github.com/bitwarden/clients/commit/ec90c01b2b922e550d02f0c1d1fdcac5621a57d6 (which is compatible with our `v2024.5.0.patch` file as far as I've tested it).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,8 @@ USER node
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2024.5.0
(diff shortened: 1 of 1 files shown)


tessus:
v2024.6.1 is out

stefan0xC:
I'm aware.  However it has a few changes that I've not yet had the time to check and also if we move the repos to the @vaultwarden organization the URL in this patch file will have to be updated as well, so I was kinda waiting for that. (If it takes a while I can create a PR or update this one and can ask @BlackDex to update #164 accordingly)

BlackDex:
I think a lot might be changed in 6.0, unless that is all behind a feature flag. Let's update it here, and I'll adjust my PR to match.

BlackDex (approved):

PR #170, opened 2024-07-05
Strip query params when resolving subpath

Timshel:
Hey

Had some issue with `subpath` handling with SSO.
Part of it was during redirection the url included query parameters and no hash mark.

I believe it makes sense to include it to make the base url resolution more robust.

### DIFF
--- a/patches/v2024.5.0.patch
+++ b/patches/v2024.5.0.patch
@@ -554,13 +554,13 @@ index c2eb37eea5..2b5ac93392 100644
 +    //
 +    // - `https://vaultwarden.example.com/base/path/`
 +    // - `https://vaultwarden.example.com/base/path/#/some/route[?queryParam=...]`
++    // - `https://vaultwarden.example.com/base/path/?queryParam=...`
 +    //
 +    // We want to get to just `https://vaultwarden.example.com/base/path`.
 +    let baseUrl = this.win.location.href;
-+    baseUrl = baseUrl.replace(/#.*/, ""); // Strip off `#` and everything after.
-+    baseUrl = baseUrl.replace(/\/+$/, ""); // Trim any trailing `/` chars.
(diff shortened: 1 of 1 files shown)


patches/v2024.5.0.patch
```diff
@@ -557,8 +557,7 @@ index c2eb37eea5..2b5ac93392 100644
 +    //
 +    // We want to get to just `https://vaultwarden.example.com/base/path`.
 +    let baseUrl = this.win.location.href;
-+    baseUrl = baseUrl.replace(/#.*/, ""); // Strip off `#` and everything after.
```
jjlin, line 560: I think you could simply replace this with
```
baseUrl = baseUrl.replace(/[#?].*/, ""); // Strip off `#` or `?` and everything after.
```
The comments should also mention the form `https://vaultwarden.example.com/base/path/?queryParam=...` if that's now a possibility.

patches/v2024.5.0.patch
```diff
@@ -557,8 +557,7 @@ index c2eb37eea5..2b5ac93392 100644
...
-+    baseUrl = baseUrl.replace(/\/+$/, ""); // Trim any trailing `/` chars.
++    baseUrl = baseUrl.replace(/(\/+|\/*#.*|\/*\?.*)$/, ""); // Strip off trailing `/`, `#`, `?` and everything after.
```
jjlin, line 560: I think it's way harder to understand what this is doing

jjlin (changes requested):

patches/v2024.5.0.patch
Timshel: Don´t know with the added comment each `|` block match a given example.
And personally I prefer each part matching fully instead of relying on fall through.

Timshel: But anyway with the PR for `v2024.6.3` already open, if accepted the easiest way to integrate it is probably for @stefan0xC to just add it, so he'll choose ;)

BlackDex:
This seems to be working just fine :)

PR #171, opened 2024-07-05
disable org seat limit validation

stefan0xC:
remove the limit of how many users can be invited to an organization.
cf. https://github.com/dani-garcia/vaultwarden/discussions/4707


BlackDex:
@stefan0xC, is this already implemented into your 2024.6 update?

BlackDex:
Ah, never mind, i see it already.
I'm probably going to decline this PR and merge the new one, unless there are pending issues or changes needed to be done on the server side for the new version.

BlackDex (approved):

PR #172, opened 2024-07-09
disable max user check

stefan0xC:
allow invitations even if there are already exactly 10 users


stefan0xC:
Ah, true. I've not thought about the `/api/plans` endpoint for so long but we probably want to set those values.

BlackDex:
Yea, looking at the new types, we are currently at 5, which is very very old.
The latest one is `20`, which is `Enterprise (Annually)`, which better matches i think.

stefan0xC:
Closing this because after talking with @BlackDex it's not something that should be fixed by patching the web-vault but by correcting the values that are returned by the server: https://github.com/dani-garcia/vaultwarden/blob/d04b94b77d95a9968f61f71969f65aaaa8a496d6/src/db/models/organization.rs#L372

PR #173, opened 2024-07-11
Add separate build script

Timshel:
Hey

A change to move part of the build process out of the docker file in an external script.

As for why:
 - It can help with testing by allowing to run locally, but this is limited by the rather strict node version checking of `npm ci`. 
 - It's mainly for the Playwright [PR](https://github.com/dani-garcia/vaultwarden/pull/4369) in Vaultwarden to be able to [build](https://github.com/Timshel/vaultwarden/blob/feature/playwright/playwright/compose/vaultwarden/build.sh#L18) aany commit without having to work with the `bw_web_builds/Dockerfile`.
 
 Additionally, removed the `chown node` logic since I had no issues when running the build.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 builds
 docker_builds
 container_builds
+vault
 web-vault
 
 # Documentation
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 builds/
 docker_builds/
 container_builds/
+vault/
 web-vault/
 
 # Other
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,48 +18,27 @@
 FROM node:18-bookworm as build
 RUN node --version && npm --version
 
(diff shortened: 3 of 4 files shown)


Timshel:
Sorry was working on the Playwright PR and did not take time to reread the `bw_web_builds` docs :(.
And since the `build_web_vault.sh` was not used in the Dockerfile did not realise it would work.

BlackDex:
Ah, no prob. Glad this can be solved with the current script already :).
It has been Wednesday already ;).

Timshel:
Are you aware of any reason why the `checkout_web_vault` and `build_web_vault` are not usedf in the Docker image ?

They are more verbose and might make more check but since the intermediate image is discarded anyway should not have much impact ?

BlackDex:
Um, there was a reason, not sure from the top of my head right now.
I would have to take a better look, but there was a good reason i think (if not, then we could change that too of course)

PR #174, opened 2024-07-11
Use checkout and build scripts in docker

Timshel:
Hey

Earlier opened a useless (sorry) PR since I had not RTFM :smile: and the docker build is not using all the available scripts.
Since I had already made the modification switching to using `checkout_web_vault` and `build_web_vault` was trivial.

Additionally, removed the `chown node` logic since I had no issues when running the build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,48 +18,28 @@
 FROM node:18-bookworm as build
 RUN node --version && npm --version
 
-# Prepare the folder to enable non-root, otherwise npm will refuse to run the postinstall
-RUN mkdir /vault
-RUN chown node:node /vault
-USER node
-
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
 # Using https://github.com/bitwarden/clients/releases/tag/web-v2024.5.1
 ARG VAULT_VERSION=9823f69c9d17e2d94de1cc005e01202dd95f0647
+ENV VAULT_VERSION=$VAULT_VERSION
+ENV VAULT_FOLDER=bw_clients
+ENV CHECKOUT_TAGS=false
 
-WORKDIR /vault
-RUN git -c init.defaultBranch=main init && \
-    git remote add origin https://github.com/bitwarden/clients.git && \
-    git fetch --depth 1 origin "${VAULT_VERSION}" && \
-    git -c advice.detachedHead=false checkout FETCH_HEAD
+RUN mkdir /bw_web_builds
+WORKDIR /bw_web_builds
 
-COPY --chown=node:node patches /patches
-COPY --chown=node:node resources /resources
-COPY --chown=node:node scripts/apply_patches.sh /apply_patches.sh
+COPY patches ./patches
+COPY resources ./resources
+COPY scripts ./scripts
 
-RUN bash /apply_patches.sh
+RUN ./scripts/checkout_web_vault.sh
+RUN ./scripts/patch_web_vault.sh
+RUN ./scripts/build_web_vault.sh
+RUN mv "${VAULT_FOLDER}/apps/web/build" ./web-vault
 
-# Build
-RUN npm ci
-
-# Switch to the web apps folder
-WORKDIR /vault/apps/web
-
-RUN npm run dist:oss:selfhost
-
-RUN printf '{"version":"%s"}' \
-      $(git -c 'versionsort.suffix=-' ls-remote --tags --refs --sort='v:refname' https://github.com/dani-garcia/bw_web_builds.git 'v*' | tail -n1 | grep -Eo '[^\/v]*$') \
-      > build/vw-version.json
-
-# Delete debugging map files, optional
-# RUN find build -name "*.map" -delete
-
-# Prepare the final archives
-RUN mv build web-vault
 RUN tar -czvf "bw_web_vault.tar.gz" web-vault --owner=0 --group=0
-
 # Output the sha256sum here so people are able to match the sha256sum from the CI with the assets and the downloaded version if needed
 RUN echo "sha256sum: $(sha256sum "bw_web_vault.tar.gz")"
 
@@ -67,7 +47,8 @@ RUN echo "sha256sum: $(sha256sum "bw_web_vault.tar.gz")"
 # The result is included both uncompressed and as a tar.gz, to be able to use it in the docker images and the github releases directly
 FROM scratch
 # hadolint ignore=DL3010
-COPY --from=build /vault/apps/web/bw_web_vault.tar.gz /bw_web_vault.tar.gz
-COPY --from=build /vault/apps/web/web-vault /web-vault
+COPY --from=build /bw_web_builds/bw_web_vault.tar.gz /bw_web_vault.tar.gz
+COPY --from=build /bw_web_builds/web-vault /web-vault
+
 # Added so docker create works, can't actually run a scratch image
 CMD [""]
--- a/scripts/.script_env
+++ b/scripts/.script_env
@@ -2,7 +2,8 @@
 # shellcheck disable=SC2034
 set -o pipefail -o errexit
 
-VAULT_FOLDER=web-vault
+VAULT_FOLDER=${VAULT_FOLDER:=web-vault}
+CHECKOUT_TAGS=${CHECKOUT_TAGS:=true}
 OUTPUT_FOLDER=builds
 
 function get_web_vault_version {
--- a/scripts/checkout_web_vault.sh
+++ b/scripts/checkout_web_vault.sh
@@ -26,6 +26,7 @@ if [[ "${VAULT_VERSION}" =~ ^20[0-9]{2}\.[0-9]{1,2}.[0-9]{1} ]]; then
 elif [[ "${VAULT_VERSION}" =~ ^v20[0-9]{2}\.[0-9]{1,2}.[0-9]{1} ]]; then
     VAULT_VERSION="web-${VAULT_VERSION}"
 fi
+
 echo "Using: '${VAULT_VERSION}' to checkout bitwarden/client."
 
 if [ ! -d "${VAULT_FOLDER}" ]; then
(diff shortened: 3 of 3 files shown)


scripts/.script_env
```diff
@@ -2,7 +2,7 @@
 # shellcheck disable=SC2034
 set -o pipefail -o errexit
 
-VAULT_FOLDER=web-vault
+VAULT_FOLDER=${VAULT_FOLDER:-web-vault}
```
BlackDex, line 5: I think it's better to use `:=` instead of `:-`.
Because `:=` will also use the default value if it's set to an empty value for example.

Dockerfile
```diff
@@ -18,56 +18,38 @@
...
-
-# Prepare the final archives
-RUN mv build web-vault
-RUN tar -czvf "bw_web_vault.tar.gz" web-vault --owner=0 --group=0
+RUN ./scripts/checkout_web_vault.sh; \
+	./scripts/patch_web_vault.sh; \
+	./scripts/build_web_vault.sh; \
+	mv $VAULT_FOLDER/apps/web/build ./web-vault
 
+RUN tar -czvf "bw_web_vault.tar.gz" ./web-vault
```
BlackDex, line 43: I would suggest to still keep the `owner` and `group` arguments for tar. Main reason is that people could running this as a different user maybe. And this at least ensures all files will have the correct user and group within the whole file

BlackDex (changes requested):

Dockerfile
Timshel: I believe the `0` was needed to set it to `root` since it was running as `node`.

BlackDex: People can still run it as non privileged use like with podman which might cause it to have a different user and group as expected.

stefan0xC:
I think we should not checkout all tags then as it takes significantly longer without any benefit to our build process: https://github.com/dani-garcia/bw_web_builds/blob/26c122f20aeb921cb22eef7dfc3add6d20ed3582/scripts/checkout_web_vault.sh#L53

BlackDex:
I'm unsure why we kept it there actually.
Maybe because if an already checked out repo exist?

Wonder what it does if you checkout an older version, and after that a newer one, how that will go

stefan0xC:
I mean, if we remove it, we'll have to rethink how the `get_web_vault_version` works or more specifically how `generate_patch_file.sh` knows what patch file to create. Also `apply_patches.sh` will use the information to guess the patch file it should use (for docker builds it might be using the latest patch file anyway)?

BlackDex:
But the Dockerfile didn't do it either, and that worked.

stefan0xC:
> But the Dockerfile didn't do it either, and that worked.

Yeah, I mean it's a different use case. (Personally I'm not using the scripts from the `master` branch to generate my patch files, so it won't matter that much but we'd probably have to update the documentation, that if one is using the `Makefile` / scripts that they'll have to manually checkout out the tags if we were to remove it.)

BlackDex:
I do use them, but not sure if it is really needed.

stefan0xC:
Easiest fix would probably be to change it to something like
```bash
git fetch --depth 1 origin ${CHECKOUT_TAG} "${VAULT_VERSION}" 
```

and set `CHECKOUT_TAG=tag` here https://github.com/dani-garcia/bw_web_builds/blob/26c122f20aeb921cb22eef7dfc3add6d20ed3582/scripts/checkout_web_vault.sh#L24-L28

Timshel:
I'll integrate it :)

Timshel:
@stefan0xC Hey was a bit unsure where you wanted the var to be set.

Decided to go with a boolean `CHECKOUT_TAGS` with a default `true` value in the `.script_env`.
At first tested the version using an empty var and `=` instead of `:=` to not use the default when empty but decided to go with the boolean version since it's more intuitive.

stefan0xC:
My proposal was not to use `--tags` at all and just checkout the specific tag if a tag is requested (which would be detected if either the `if` or the `elif` condition is true).

Timshel:
O, I understand will make the change

Timshel:
Ok, hope this is right now :smile:, if it's not the case sorry and should have let you do the change :sweat_smile:.

Dockerfile
```diff
@@ -18,56 +18,38 @@
...
-COPY --chown=node:node scripts/apply_patches.sh /apply_patches.sh
+COPY patches ./patches
+COPY resources ./resources
+COPY scripts ./scripts
 
-RUN bash /apply_patches.sh
+RUN ./scripts/checkout_web_vault.sh; \
+	./scripts/patch_web_vault.sh; \
+	./scripts/build_web_vault.sh; \
+	mv $VAULT_FOLDER/apps/web/build ./web-vault
```
stefan0xC, line 41:
```suggestion
RUN ./scripts/checkout_web_vault.sh
RUN ./scripts/patch_web_vault.sh
RUN ./scripts/build_web_vault.sh
RUN mv $VAULT_FOLDER/apps/web/build ./web-vault
```

I think if you separate the statements docker can cache the intermediate results as separate layer.

Timshel: I'm guessing when building locally in case of error ?
Otherwise, in general more layers are not exactly a plus, no ?

BlackDex: In this case it doesn't matter for the end result. We use a from scratch there. So everything before that will not end up into the final image. 

And having separate runs here is better indeed. 
Good catch.

scripts/checkout_web_vault.sh
BlackDex, line 1: This unfortunately also does not work since we checkout a git commit hash, and not a specific tagged version, this will also fail, the tag will not be fetched in this case and thus will not be able to fetch it.

I was looking around a bit regarding this, and I found the following option to use, which seems to work for me.

If it is a valid sha1 commit hash
```bash
if [[ ${#VAULT_VERSION} -eq 40 && "${VAULT_VERSION}" =~ ^[a-f0-9]+$ ]]; then
  git clone -c "remote.origin.fetch=+${VAULT_VERSION}:refs/remotes/origin/${VAULT_VERSION}" https://github.com/bitwarden/clients.git --depth 1 "${VAULT_FOLDER}"
else
  git clone --branch "${VAULT_VERSION}" --depth 1 "${VAULT_FOLDER}"
fi
```

The only thing i do not know is, how does this work if the repo is already checked-out and we do a git stash etc.. etc.. 

But this does work great on a first checkout.
Maybe we need to move an existing folder to rename it to `old_` or something, and then also clean those during `make clean` or something?

BlackDex (changes requested):

scripts/checkout_web_vault.sh
Timshel: I'm not sure I follow what is not working ?

BlackDex: Determining the tag/branch of the checkout when using a sha. That breaks.

BlackDex: No, not only that. With the code i provide, we do not need tags at all. All would be working just fine and quicker.

BlackDex: Ok. Did some checking. But it looks like this is going to give issues if there already is a previous directory.

Now, from my point of few we can do multiple things.
1. Leave it as is, in the sense that we keep the `--tags` and take for granted it will take longer to download.
2. Always delete the web-vault directory, but generate a patch from the `stash` action
    So, after the `git stash` to stash all run `git stash show -p > to/some/file.patch`

I'm leaning towards the latter. As that makes it all a bit more clean and still faster.
What do you think?

Timshel: I would stay with the first solution since :
 - It's what is currently done, so it's more inline with the goal of the PR to use the scripts in docker
 - The stash logic look a bit cumbersome to me
 - The perf impact is quite low since if the folder already exists then you already fetched most of the tags which mean there will be almost no difference later on.

scripts/checkout_web_vault.sh
```diff
@@ -47,10 +48,15 @@ else
...
+CHECKOUT_ARGS=""
+if [ "$CHECKOUT_TAGS" = "true" ]; then
+    CHECKOUT_ARGS="$CHECKOUT_ARGS --tags"
+fi
+
 # Checkout the request
 pushd "${VAULT_FOLDER}"
     # Update branch and tag metadata
-    git fetch --tags --depth 1 origin "${VAULT_VERSION}"
+    git fetch --depth 1 origin ${CHECKOUT_ARGS} "${VAULT_VERSION}"
```
BlackDex, line 59: Some nitpicking here to fix some shellcheck linting and how the command is buildup.

The `CHECKOUT_ARGS` is probably better to be an array. And using `CHECKOUT_ARGS` within it self is not needed, as nothing else will be able to have something added before.

Also, all options should be added before `origin` or any other specific repo assignment.


```suggestion
CHECKOUT_ARGS=()
if [ "${CHECKOUT_TAGS}" = "true" ]; then
    CHECKOUT_ARGS+=(--tags)
fi

# Checkout the request
pushd "${VAULT_FOLDER}"
    # Update branch and tag metadata
    git fetch "${CHECKOUT_ARGS[@]}" --depth 1 origin "${VAULT_VERSION}"
```

BlackDex (changes requested):

scripts/checkout_web_vault.sh
Timshel: I set it within itself as an easy way to inject additional args if needed, forgot to add the `:-` in case `set -u` is ever used.  Switching to an array make injection impossible and things more complicated.

Will fix the order.

Dockerfile
```diff
@@ -18,56 +18,37 @@
...
-WORKDIR /vault/apps/web
+COPY patches ./patches
+COPY resources ./resources
+COPY scripts ./scripts
 
-RUN npm run dist:oss:selfhost
+RUN ./scripts/checkout_web_vault.sh
+RUN	./scripts/patch_web_vault.sh
+RUN	./scripts/build_web_vault.sh
+RUN	mv $VAULT_FOLDER/apps/web/build ./web-vault
```
dfunkt, line 40: The spacing here doesn't look right.

BlackDex: which spacing?
Looks fine too me.

The only remark i might have is that it might be better to use `mv "${VAULT_FOLDER}/apps/web/build" ./web-vault` though.

BlackDex (changes requested):

Dockerfile
dfunkt: Just compare it with the line above the part I selected:
`RUN ./scripts/checkout_web_vault.sh`
vs
```
RUN	./scripts/patch_web_vault.sh
RUN	./scripts/build_web_vault.sh
RUN	mv $VAULT_FOLDER/apps/web/build ./web-vault
```
I think he might have use tabs instead of spaces for these 3 lines ^^

BlackDex (approved):

PR #176, opened 2024-08-08
Strip query params when resolving subpath

Timshel:
Same as https://github.com/dani-garcia/bw_web_builds/pull/170 but applied on latest patch [v2024.6.2.patch](https://github.com/dani-garcia/bw_web_builds/blob/master/patches/v2024.6.2.patch).

Edit: It appears that https://github.com/dani-garcia/bw_web_builds/pull/171 is not included too in the latest release; not sure if it's on purpose.

### DIFF
--- a/patches/v2024.6.2.patch
+++ b/patches/v2024.6.2.patch
@@ -577,11 +577,11 @@ index c2eb37eea5..2b5ac93392 100644
 +    //
 +    // - `https://vaultwarden.example.com/base/path/`
 +    // - `https://vaultwarden.example.com/base/path/#/some/route[?queryParam=...]`
++    // - `https://vaultwarden.example.com/base/path/?queryParam=...`
 +    //
 +    // We want to get to just `https://vaultwarden.example.com/base/path`.
 +    let baseUrl = this.win.location.href;
-+    baseUrl = baseUrl.replace(/#.*/, ""); // Strip off `#` and everything after.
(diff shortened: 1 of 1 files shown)


BlackDex:
I guess that is because it wasn't in the patch-file @stefan0xC created.
I also did not noticed this, or even thought about that at all yesterday.

I can merge this one right now, but I'm probably not going to release a new web-vault right away, mainly because OIDC is not yet part of the main branch and i do not like adding letters to our versions haha.
I Think that it then would be best if this is going to be part of the next web-vault version from the start.

What do you think @Timshel and @stefan0xC ?

FYI: After the next Vaultwarden release i tend to spend some more time on checking the OIDC PR.

Timshel:
No worries :).

On the release part, unless there is another one scheduled soon, I would prefer not to wait an unknown amount of time. 
I had already removed this part of the fix from the last version I packaged meaning if not released I'll have to add it back and then remove it again later on.

stefan0xC:
#171 is on purpose the other probably an oversight of mine

BlackDex:
Yes #171 is fixed server side, not needed for the client.

BlackDex:
Well, i might release a new version then anyways hehe

BlackDex:
Ok @Timshel, new release created, including Vaultwarden `testing` is being build at this moment.
If you rebase your code it should be working correctly without the need for any other patches.

Timshel:
@BlackDex thx :)

PR #169, opened 2024-06-19
update web-vault to v2024.6.2 and remove bitwarden_license/ from build

stefan0xC:
New web-vault version: [web-v2024.6.2](https://github.com/bitwarden/clients/releases/tag/web-v2024.6.2)

There were a lot of [changes between v2024.5.1 and v2024.6.2](https://github.com/bitwarden/clients/compare/web-v2024.5.1...web-v2024.6.2) 

Since the payment method warning is now hidden behind a feature flag I've removed the code that patches it out (because the banner was also added to other parts of the code base and I think the patches are more maintainable this way). We might want to remove it manually again if the feature flag gets removed.

PR #179, opened 2024-08-30
backport master password verification support

stefan0xC:
if you are using biometrics or pin to unlock your android device and then use that device to login, the user verification e.g. when exporting your personal vault will not work (cf. https://github.com/dani-garcia/vaultwarden/issues/4875)

so we also need to backport https://github.com/bitwarden/clients/commit/1043a582c1426e04917dfc0657a5deb6fe5784cb to add server side master password verification support (https://github.com/bitwarden/clients/pull/9523) for the `v2024.6.2.patch`

### DIFF
--- a/patches/v2024.6.2.patch
+++ b/patches/v2024.6.2.patch
@@ -865,6 +865,125 @@ index f4e97262a3..12e11b35a3 100644
  
      "<rootDir>/libs/admin-console/jest.config.js",
      "<rootDir>/libs/angular/jest.config.js",
+diff --git a/libs/angular/src/auth/components/lock.component.ts b/libs/angular/src/auth/components/lock.component.ts
+index 64cd664f1f..88b042c5b8 100644
+--- a/libs/angular/src/auth/components/lock.component.ts
++++ b/libs/angular/src/auth/components/lock.component.ts
+@@ -17,9 +17,12 @@ import { KdfConfigService } from "@bitwarden/common/auth/abstractions/kdf-config
+ import { InternalMasterPasswordServiceAbstraction } from "@bitwarden/common/auth/abstractions/master-password.service.abstraction";
+ import { UserVerificationService } from "@bitwarden/common/auth/abstractions/user-verification/user-verification.service.abstraction";
+ import { AuthenticationStatus } from "@bitwarden/common/auth/enums/authentication-status";
++import { VerificationType } from "@bitwarden/common/auth/enums/verification-type";
+ import { ForceSetPasswordReason } from "@bitwarden/common/auth/models/domain/force-set-password-reason";
+-import { SecretVerificationRequest } from "@bitwarden/common/auth/models/request/secret-verification.request";
+-import { MasterPasswordPolicyResponse } from "@bitwarden/common/auth/models/response/master-password-policy.response";
++import {
++  MasterPasswordVerification,
++  MasterPasswordVerificationResponse,
++} from "@bitwarden/common/auth/types/verification";
+ import { VaultTimeoutAction } from "@bitwarden/common/enums/vault-timeout-action.enum";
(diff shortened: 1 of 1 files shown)


stefan0xC:
patch files for future web-vaults should be less messy again as we don't need to backport this fixes. 

the alternative to backporting those two commits would be to temporarily disable the master password verification for the `v2024.6.2` web-vault as suggested here https://github.com/dani-garcia/vaultwarden/issues/4628#issuecomment-2158830814)

BlackDex:
Either that, or we need to somehow add extra ways to add some cherry-picked patches.

But, lets first see if we can get a newer vault working at all.

stefan0xC:
Well, I am kinda waiting on https://github.com/bitwarden/clients/pull/10549 to be merged and released. I mean, we could also use the `tw-fill-text-headers` class as a workaround for now as suggested in https://github.com/dani-garcia/bw_web_builds/pull/169#issuecomment-2209905446 but I have not found the time or motivation to work on this at the moment. (And the other issue https://github.com/dani-garcia/bw_web_builds/pull/169#issuecomment-2239478617 is maybe not that big of a deal either but we probably want to use SVG instead of PNG for the logo.)

> we need to somehow add extra ways to add some cherry-picked patches.

Since I'm using my soft fork approach to generate the patch file it's not a problem for me personally because I can just cherry-pick from future branches and adapt them as needed (cf. https://github.com/bitwarden/clients/compare/web-v2024.6.2...stefan0xC:bitwarden_clients:v2024.6.2). And this specific commit includes changes to a not-yet-existing file for the CLI ([apps/cli/src/oss-serve-configurator.ts](https://github.com/bitwarden/clients/blob/web-v2024.7.1/apps/cli/src/oss-serve-configurator.ts)) so we would have to ignore that too if we use another way to add cherry-picked patches. (I mean we could store them as separate patch files but what would be the benefit to the current approach? We would still have to apply them the same.)

BlackDex:
With the soft-fork you could also have that as a sub-repo and maybe a shallow fork? Excluding specific directories not not being pulled at all?

I still have to see in more detail how the soft-fork is that much of a difference in the sense that the patches made and then doing a pull or whatever makes it difficult to maintain and update.

But i have not looked into it really good. If you have some (detailed) steps then i can try it my self.

Also, that way of working would be better for the vaultwarden GitHub project to be used instead of modifying this one 😄

stefan0xC:
> With the soft-fork you could also have that as a sub-repo and maybe a shallow fork? Excluding specific directories not not being pulled at all?

Maybe? I mean I was not really concerned with that. To me the benefit of my current approach is that I keep track of each web-vault version in a separate branch (as fork of the bitwarden/clients repo). I mean I have added this `bitwarden_clients` repo for testing purposes as a sub module to my `bw_web_builds` fork (which is mainly used to hold my scripts to [update to the newest web-vault](https://github.com/stefan0xC/bw_web_builds/blob/vaultwarden-web/scripts/prepare_update.sh) and create a compatible patch file to this repo) but like I said that is only for testing purposes (and I agree that this would be something for the vaultwarden organization).

My workflow in general is very straight-forward: fetching the latest web-vault release, creating a new branch from the tag (without the `web-` prefix) and then I `git cherry-pick` the changes from a previous release (e.g. `web-v2024.7.1..v2024.7.1`). While cherry-picking the changes for Vaultwarden I will have to fix each merge conflict when it comes up - so this still needs to be done but in my opinion it's much more convenient doing it interactively this way then manually applying the changes that could not be applied.

To [create a compatible patch file for this repository](https://github.com/stefan0xC/bw_web_builds/blob/6f99670c6fafe803f1fb0c523c24e1a53cfca2e2/scripts/create-patch-file.sh#L33-L55) I ignore the changes to certain files and directories (e.g. the icons, logos and images).

BlackDex (approved):

PR #182, opened 2024-09-18
update web-vault to v2024.10.5

stefan0xC:
Added two SVGs for the login page (and adapted it a bit so our logo is a bit bigger and the page is not missing the logo when logging out)...
The only thing missing at the moment is replacing the bitwarden shield when the side panel is collapsed:

https://github.com/user-attachments/assets/382630fc-30a1-4151-aa39-81e04c3dd129

And I also have not checked what server-side changes are required.


BlackDex:
How about manager roles?
I think that would be the biggest impediment here.

ecesarini:
Hi,
could you help me to understand how the new svg logos (in the log in page)  are handled in the code? I mean the whole part above the login form.
Until v2024.6.2 there were html tags in apps/web/src/app/auth/login/login.component.html, e.g.
```
    <div>
      <img class="logo logo-themed" alt="Bitwarden" />
      <p class="tw-mx-4 tw-mb-4 tw-mt-3 tw-text-center tw-text-xl">
        {{ "loginOrCreateNewAccount" | i18n }}
      </p>
```
In the following releases those have been deleted.
I use to add a further logo above the login form before applying the vaultwarden bw-web patch, and I'm trying to understand how to hook the upstream code.

Thank you very much

stefan0xC:
I can post screenshots later and explain the differences but you can also look at https://vault.bitwarden.com how it will look with the logos (e.g. where they are placed).

ecesarini:
> I can post screenshots later and explain the differences but you can also look at https://vault.bitwarden.com how it will look with the Bitwarden logos (e.g. where they are placed).

Hi,
I've already taken a look there but it didn't help me. I would like to place some html content above the new bw-shield (that is the new vultwarden-icon i think) or above the "Log in" string but i can't find the angular project file involved in that.

If you can give me some clues later, it would be great.

Thank you again

stefan0xC:
@ecesarini okay, so there are 3 svg files that I've changed:
[`apps/web/src/images/logo.svg`](https://github.com/bitwarden/clients/blob/web-v2024.9.1/apps/web/src/images/logo.svg) should be the same as the `BitwardenLogo` embedded here: [`libs/auth/src/angular/icons/bitwarden-logo.icon.ts`](https://github.com/bitwarden/clients/blob/web-v2024.9.1/libs/auth/src/angular/icons/bitwarden-logo.icon.ts)
The first file is [loaded via css](https://github.com/bitwarden/clients/blob/601c13ae1b5ea45c1a47426c9adf190a14ebbd59/apps/web/src/scss/tailwind.css#L58) in the [index.html](https://github.com/bitwarden/clients/blob/web-v2024.9.1/apps/web/src/index.html#L19) (but this is only shown before the app is loaded or if you have disabled javascript) and the [second logo is referenced here](https://github.com/bitwarden/clients/blob/web-v2024.9.1/libs/auth/src/angular/anon-layout/anon-layout.component.html#L9) (when you are not logged in, while the vault is locked or when viewing a send).

the other change is the Bitwarden shield logo embedded in [`libs/auth/src/angular/icons/bitwarden-shield.icon.ts`](https://github.com/bitwarden/clients/blob/web-v2024.9.1/libs/auth/src/angular/icons/bitwarden-shield.icon.ts) which is shown [above the Log In](https://github.com/bitwarden/clients/blob/601c13ae1b5ea45c1a47426c9adf190a14ebbd59/libs/auth/src/angular/anon-layout/anon-layout.component.html#L13) (also only when not logged in).

Light mode:
Screen Shot 2024-09-18 at 20 11 25
Dark mode:
Screen Shot 2024-09-18 at 20 11 13

Note that I'm using the new `class="tw-fill-marketing-logo"` for the appropriate dark / light mode fill color and not a separate [`logo-white.svg`](https://github.com/bitwarden/clients/blob/web-v2024.9.1/apps/web/src/images/logo-white.svg) file.

The shield icon will be replaced by an icon of a padlock when the vault is locked.

stefan0xC:
> How about manager roles? I think that would be the biggest impediment here.

The Manager role (as well as the access to all collections options) has been removed so users will be listed as User:
image

BlackDex:
Well if this works, it is at least a good starting point for testing this.

stefan0xC:
> Note that I'm using the new `class="tw-fill-marketing-logo"` for the appropriate dark / light mode fill color and not a separate [`logo-white.svg`](https://github.com/bitwarden/clients/blob/web-v2024.9.1/apps/web/src/images/logo-white.svg) file.

Just noticed that this does not seem to work because the css style is not applied to the referenced SVG. :neutral_face:

Gerardv514:
What is the status/obstacles facing with this web build?

BlackDex:
The manage role is not supported by Vaultwarden yet. Without that we can't use this web-vault version.

We either need to internally convert access-all to manage and vice-versa without making to much change into the database structure or add role support fully, which takes longer to add.

The conversion option has a lot of potential to break stuff i think, so probably not the best option.

stefan0xC:
I've updated the PR to the latest web-vault ([v2024.10.5](https://github.com/bitwarden/clients/releases/tag/web-v2024.10.5)).

Note: I've also updated the bitwarden/sdk-internal dependency to the GPL licensed [version 0.17.0](https://www.npmjs.com/package/@bitwarden/sdk-internal/v/0.1.7). As far as I know this should not have any influence as the code is not even used in the building of the web-vault but this means that the version we would ship would deviate in this regard from the bitwarden version.

BlackDex:
Thanks. Ill use this version to test/develop the manager-role removal changes.

stefan0xC:
closing in favor of #186

PR #183, opened 2024-09-29
Reached heap limit Allocation on webpack build

alebo-iX:
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

### DIFF
--- a/scripts/build_web_vault.sh
+++ b/scripts/build_web_vault.sh
@@ -16,6 +16,8 @@ trap 'handle_error $LINENO $?' ERR
 # Show used versions
 node --version
 npm --version
+#set virtual memory allocation
+export NODE_OPTIONS=--max-old-space-size=4096
 
 # Build
 pushd "${VAULT_FOLDER}"


maurice-audin:
+1

Build on CI is broken because of this since v2024.5.0

BlackDex:
Our CI works just fine: https://github.com/dani-garcia/bw_web_builds/actions/workflows/release.yml

Using the docker build also works just fine.
My understanding is that some systems need this because either there systems does not have a lot of memory available, or the node version is a bit older and has some compatibility issues with the system it is running on.

If this is within a CI, you can of course also just add that variable to your CI env's i would imagine.
It might have adverse effects maybe on systems where it runs fine without this option set.

BlackDex:
My main issue is, that this setting could cause issues on systems which have less memory, where for some reason the build works just fine. Adding this by default might cause issues there.

I can think of other ways to solve this though, but that would mean that we would add something like a `.env_build` or something which we then load in the needed scripts. That way it is more flexible.

maurice-audin:
A single line change would be in the Makefile:

```
$ git diff Makefile
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ full: checkout patch-web-vault build tar
 .PHONY: full
 
 container:
- ${CONTAINER_CMD} build -t bw_web_vault .
+ ${CONTAINER_CMD} build ${CONTAINER_BUILD_ARGS} -t bw_web_vault .
 .PHONY: container
 
 container-extract: container
```

We could then set the variable CONTAINER_BUILD_ARGS to "--build-arg NODE_OPTIONS=--max-old-space-size=4096" in CI environment and behavior would stay the same with an empty variable.

BlackDex:
Using a build_env file is more robust I think.

BlackDex:
@alebo-iX i have created a PR which should solve this issue in a more flexible way, instead of forcing the memory option on all builds, or add complex and error prone variables which might be empty or not, or contain globs etc...

See #184

PR #186, opened 2024-11-21
Update to v2024.12 and Dynamic CSS

BlackDex:
Update web-vault to v2024.11
- This PR includes the changes from @stefan0xC done in #182
- It also adds the dynamic-css support
- Added a CI job which runs `make docker-extract` and uploads the generated `.tar.gz` as an artifact.

Removed the access-intelligence since this is linked to Bitwarden Licensed code.


BlackDex:
@stefan0xC i based this PR upon yours and updated to v2024.11.1 and added dynamic-css here.
The web-vault built using this PR can be used with https://github.com/dani-garcia/vaultwarden/pull/5219

It would be nice if people are able to test this web-vault with the code in the PR at the Vaultwarden repo.
And then mainly the whole manager role and access-all etc...

BlackDex:
This PR using v2024.11.1 web-vault version only works with this pending PR https://github.com/dani-garcia/vaultwarden/pull/5219 .

Several changes were needed on the server side to use the Custom Role as a Manger role and still be compatible with a previous version of Vaultwarden in case someone needs/wants to rollback. So do not try to use this PR with any version equal to or older then v1.32.5, it will not work.

scripts/apply_patches.sh
```diff
@@ -48,9 +48,16 @@ replace_embedded_svg_icon \
...
+replace_embedded_svg_icon \
+	../resources/src/images/logo.svg \
+	./libs/auth/src/angular/icons/bitwarden-logo.icon.ts
+replace_embedded_svg_icon \
+	../resources/vaultwarden-icon.svg \
+	./libs/auth/src/angular/icons/bitwarden-shield.icon.ts
 
 echo "Remove non-free bitwarden_license/ code"
 rm -rf ./bitwarden_license/
+rm -rf ./apps/web/src/app/tools/access-intelligence/
```
dani-garcia, line 60: Note, this should be resolved with next version, so we can remove this by then: https://github.com/bitwarden/clients/pull/12116

tessus: How about this then?

```suggestion
[ -d ./apps/web/src/app/tools/access-intelligence/ ] && rm -rf ./apps/web/src/app/tools/access-intelligence/
```

BlackDex: Yea, that seems like a good one.
Though, the patch needs to be fixed to, it will at least not break on this point then already.

BlackDex:
I Updated to the just released v2024.11.2 security patch.

BlackDex:
I'm checking to see if i can update to v2024.12

BlackDex:
Ok, i have updated it to use v2024.12.0. Not too much was changed. A file was relocated and some colors have changed a bit.

Also needed to make some changes on the server-side for this, but not too much either.

dfunkt:
Is there any point in keeping the v2024.10.5 and v2024.11.2 patch files since they're now outdated/not used in any build?

BlackDex:
Might be useful for historical usage or comparing  what was changed.
It also doesn't harm to have them.

dani-garcia (approved):

PR #189, opened 2025-01-22
update web-vault to v2025.1.1

stefan0xC:
New web-vault release: [web-v2025.1.1](https://github.com/bitwarden/clients/releases/tag/web-v2025.1.1)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2025.1.0
-ARG VAULT_VERSION=4fb35dbdf11dc781ffe6cdef1173f821fa6ebeb3
+# Using https://github.com/bitwarden/clients/releases/tag/web-v2025.1.1
+ARG VAULT_VERSION=06b60c6853dc6c510eaf5ee68d547d02baee8934
 ENV VAULT_VERSION=$VAULT_VERSION
 ENV VAULT_FOLDER=bw_clients
 ENV CHECKOUT_TAGS=false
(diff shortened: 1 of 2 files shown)


BlackDex:
There is a new `Devices` tab in the user settings.
That needs at least an endpoint for show the devices, and probably also some extra's to probably remove access.

I wanted to add something similar to the Vaultwarden Admin environment, but have not yet come around to do that.

BlackDex (approved):

tessus:
@BlackDex is it still necessary for the web vault to be marked as pre-release?

BlackDex:
Yes, since there is no stable release yet of Vaultwarden it self which includes this version. This way we prevent users or packagers from trying to use a web-vault with an older release of Vaultwarden.

tessus:
Hmm, I see. Thanks.

Although just as an info, packagers do not always package only releases. e.g. I create packages in the form x.y.z^YYYYMMDD.githash as well. So do a lot of others.

But you could make the mesaage a bit less restrictive:

DO NOT USE THIS WITH ANY STABLE OR LATEST VAULTWARDEN YET!

should rather be:

DO NOT USE THIS WITH ANY STABLE VAULTWARDEN YET!

stefan0xC:
If packagers want to ship a version based of Vaultwarden's `main` branch instead of waiting for a release they can do so. Not sure why this pre-release message of the web-vault has to be changed for that.

tessus:
This message sounds like the web vault will destroy your data (YOU HAVE BEEN WARNED) if used with the testing branch. 

Either way this sentence is ambiguous and ahould be clarified.

You might know that `latest` means the latest release in this project. For most people however `latest` means the latest code, which in turn means the most recent commit in the main/dev/master branch (whatever branch is used to cut releases).

So how about:

DO NOT USE THIS WITH THE LATEST RELEASE OF VAULTWARDEN YET!

BlackDex:
We can discuss the terms latest, testing, beta, alpha etc...  But there are more projects which use `latest` as just a tag for there stable releases, and `beta` or `dev` for there _rolling_ releases.

For this projects, `stable` and `latest` are the same, and how do you know for sure it might not break your whole vault if using this newer version of the web-vault with an older Vaultwarden, it could be a possibility.

I'm going to convert it to release once there is a new version of Vaultwarden, not earlier, since we had reports in the past of people using newer web-vault's with older versions and then complaining stuff didn't worked.

Issue #137, opened 2023-08-14
Rethinking the web-vault repository


stefan0xC:
I've been working on a [soft fork of the clients repository](https://github.com/stefan0xC/bitwarden_clients) where the changes to the web-vault can be directly applied. The idea is to having a separate branch for each web-vault release, which is likely easier to manage than the way we currently do with the patch files. And we could manage the releases using that repository as a submodule in this repository [like this](https://github.com/stefan0xC/bw_web_builds/tree/vaultwarden-web).

While this is a proof of concept right now, I believe it would make it easier to keep track of the changes in the future because the changes can be annotated in the commits (cf. my [v2023.7.1 branch](https://github.com/bitwarden/clients/compare/web-v2023.7.1...stefan0xC:bitwarden_clients:v2023.7.1)).

If we want we could also add the old patch files more or less unmodified like I've done in [v2023.5.1.patch](https://github.com/bitwarden/clients/compare/web-v2023.5.1...stefan0xC:bitwarden_clients:v2023.5.1).

By using something like `git cherry-pick v2023.7.1~4..v2023.7.1` (which I've tested on the `rc` branch) it's easy to re-apply the changeset to a new release (and it's also more comfortable to use in case of merge conflicts).

Keeping the changes to the [bitwarden/clients repository](https://github.com/bitwarden/clients/) separate from this repository would also make it possible to just change the `version` in `app/web/package.json` so we could get rid of the `vw-version.json` file entirely.

The obvious drawback is that we'd have to maintain a fork but I think that it would be worth it.

BlackDex:
I think this way of working has been mentioned a few times before.
I do see some issues with this. mainly that it could be harder for other people if the want to pull the bitwarden client repo them selfs and patch it for Vaultwarden.

I also think the patch files make it a bit more transparent into what is going to be patched.
If i understand the way you mention correctly, is that we just fork the client repo, and make the adjustments and to a rebase everytime needed right?

That might be easier for small changes and sometimes also for some larger, but i think it will loose a good way to see what has been patched. specifically for Vaultwarden.

It might be that i overlook something, so please let me know, or explain it better or even graphically explain it which might help others also to understand it.

stefan0xC:
[...]

PR #191, opened 2025-03-05
Build using prepatched vaultwarden/vw_web_builds

Timshel:
In reference of https://github.com/dani-garcia/bw_web_builds/issues/137

Build the project using an already patched [vaultwarden/vw_web_builds](https://github.com/vaultwarden/vw_web_builds).

Cleanup the scripts and tested the docker build and `make full`.


stefan0xC:
The commit id `172cbd2e77e235e0eead6bb5022ad9185a45cafd` is the [web-v2025.2.2](https://github.com/bitwarden/clients/releases/tag/web-v2025.2.2) tag. I have not yet merged the changeset for the v2025.2.2 branch https://github.com/vaultwarden/vw_web_builds/pull/8 so this currently would still use the unpatched version.

Timshel:
Yes, I was aware just wanted to point to the correct branch.
Updated to https://github.com/vaultwarden/vw_web_builds/commit/31e5dd1b809139012d2092b827d59815c14e5aec which I believe is latest now.

stefan0xC:
Ah, yeah, "Rebase and merge" also changes the commit id.

Timshel:
@stefan0xC do you plan to update the [v2025.3.1](https://github.com/vaultwarden/vw_web_builds/commits/v2025.3.1) branch ?

stefan0xC:
It has been merged.

Timshel:
Updated to https://github.com/vaultwarden/vw_web_builds/commit/19f6dee8d32659ad3a26c33176301175a4d8a11e

BlackDex:
Yea, this seems like a good starting point for the short term.

I'm only missing a good way of working for the `vw_web_builds`.
Like, how do we keep in-sync with upstream.
Do we keep the `main` branch 1:1 in-sync with upstream
Should we create an own _main_ branch, like `vw_main` which we link to a specific client tagged version?
And when a new version is released by Bitwarden, we sync that branch to a specific tag/hash and apply changes if needed?
And, then tag our custom versions upon the `vw_main` branch?

It might be best if we have a document on how to work on that so that we can't make a weird mistake (Which is always possible of course).

I have never maintained a fork with my own changes and try to keep upstream patches/changes in sync. So i have no real clue. I might have a general idea though.

Timshel:
I believe the current process used @stefan0xC is working well: 

- Creating a new branch for each Bitwarden release.
- Rebasing the modifications on top of the tag.

With this process I believe preparing a new release is as simple as starting from the previous version and using an interactive rebase on top of the new tag to ensure to keep only the Vaultwarden modifications and fixing the conflicts.

This allows to easily keep track of each release and do not require to force push on a `main` branch.

Only suggestion I would add is later on to move the release code from this repository to the `main` branch of `vw_web_builds`.

This would allow to  :

- Use only one repository / archive this one.
- The main branch would then have a reference to the latest release.

stefan0xC:
I' have skipped the separate PR step for my changeset and just pushed a [new `v2025.4.1` branch](https://github.com/vaultwarden/vw_web_builds/tree/v2025.4.1) directly. If we adopt this PR we would still have to update the pointer to the branch we want to build, etc.

scripts/checkout_web_vault.sh
```diff
@@ -34,7 +34,7 @@ if [ ! -d "${VAULT_FOLDER}" ]; then
     pushd "${VAULT_FOLDER}"
         # If this is the first time, init the repo and checkout the requested branch/tag/hash
         git -c init.defaultBranch=main init
-        git remote add origin https://github.com/bitwarden/clients.git
+        git remote add origin https://github.com/vaultwarden/vw_web_builds.git
```
stefan0xC, line 37: I'd rename it vaultwarden so it does not interfere with already checked out repositories

```suggestion
        git remote add vaultwarden https://github.com/vaultwarden/vw_web_builds.git
```
(This would then have to be changed on line 58 too to fetch from vaultwarden.)

scripts/checkout_web_vault.sh
stefan0xC, line 1: The version check is not working for manually providing a web-vault version (e.g. `VAULT_VERSION="v2025.4.1" make full`) because there are no tags with `web-` so the check would have to be reversed or just simplified to prepending a missing `v` in case of YYYY.M.B. (For docker builds this should not matter as we are using the commit id anyway.)

Dockerfile
```diff
@@ -23,23 +23,20 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2025.3.1
-ARG VAULT_VERSION=f941cb5c29ba0af2902cd0226540378bdeba568c
+# Using https://github.com/vaultwarden/vw_web_builds/tree/v2025.3.1
+ARG VAULT_VERSION=19f6dee8d32659ad3a26c33176301175a4d8a11e
```
stefan0xC, line 27: For building the newest web-vault this should be updated to 
```suggestion
# Using https://github.com/vaultwarden/vw_web_builds/tree/v2025.4.1
ARG VAULT_VERSION=166cb75401f4578a8d809186df3ca9e1134c0113
```

scripts/generate_patch_file.sh
stefan0xC, line 1: Not sure if we should remove it right now. I think for reviewing changes it's probably good to have a way to create a new patch file  (though this script not work anymore because we would actually need to compare the changes from the web-vault tag from the bitwarden/clients repository to the branch from the vaultwarden repository). I think it's fine but pointing it out because we probably also have to update the documentation accordingly.

stefan0xC (changes requested):

scripts/generate_patch_file.sh
stefan0xC: I mean we could also add the `web-v` tags to the vw_web_builds repository to make it easier to compare as well. :thinking: 

For new changes they'd have to be made in the https://github.com/vaultwarden/vw_web_builds repository as well (unless they concern the building of the repository).

stefan0xC:
Oh, and I've just realized that I have forgotten to remove the `bitwarden_license` code from newer versions of the web-vault (ever since https://github.com/vaultwarden/vw_web_builds/pull/2 has been merged). So this would have to be rectified in our fork as well because without the `apply_patches.sh` script those files will not be removed after checkout anymore.

scripts/checkout_web_vault.sh
Timshel: Gone with the simplified verison.

Timshel:
`web-v2025.4.2` was released maybe we could try to make the release using this PR ?

BlackDex:
> `web-v2025.4.2` was released maybe we could try to make the release using this PR ?

Sounds good to me.

stefan0xC:
> `web-v2025.4.2` was released maybe we could try to make the release using this PR ?

I've added [the changeset](https://github.com/bitwarden/clients/compare/web-v2025.4.2...vaultwarden:vw_web_builds:v2025.4.2) for [`v2025.4.2`](https://github.com/vaultwarden/vw_web_builds/tree/v2025.4.2). cf. https://github.com/vaultwarden/vw_web_builds/commit/b9151b7da39d464fdd4c271e23603721890dc0fb

BlackDex:
Well, the CI Build seems to have passed and created a tar file.
So, that is nice.

stefan0xC:
I've updated the web-vault for the new [`v2025.5.0`](https://github.com/vaultwarden/vw_web_builds/tree/v2025.5.0) and noticed that since `web-2025.4.1` the create new organization screen throws an exception because `this.taxComponent` is not found. I've updated the branches to disable the sales tax accordingly:
https://github.com/vaultwarden/vw_web_builds/commit/80c6b74cecd4a309e1372d4458c61467b616acd0

BlackDex (approved):

PR #193, opened 2025-04-16
update web-vault to v2025.4.0

stefan0xC:
update to the newest web-vault: [web-v2025.4.0](https://github.com/bitwarden/clients/releases/tag/web-v2025.4.0)

also includes a [patch](https://github.com/stefan0xC/bitwarden_clients/commit/6ba495828ebc2b2a2c6482e8cd0aeb368da04843) that replaces "Log in to Bitwarden"  with just "Log in"
image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/bitwarden/clients/releases/tag/web-v2025.3.1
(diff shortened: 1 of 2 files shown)


stefan0xC:
> The major feature is probably the export with attachments, which we do not (yet) support.

As fas as I've tested it, if you enable the `export-attachments` feature flag this just works though it's not yet enabled in Bitwarden Cloud either and it will also take a bit (of both time and memory) depending on the combined size of the uploaded attachments.

BlackDex:
Yea, so we might be able to add it as an available feature-flag we could support then.

polyzen:
Are these last two releases ready for production? The releases page has them marked as pre-release.

BlackDex:
They are still pre-release.

PR #200, opened 2025-05-27
fix local build version

stefan0xC:
Since we have switched to using our own repository I think it should be fine to revert the checkout to the requested branch, tag or git hash instead, so the `tag` and `branch` info could be used for local building if requested explicitly. (However, I am not sure why we switched to `FETCH_HEAD` in the first place.)

Thanks to @kpfleming for pointing this out.

### DIFF
--- a/scripts/checkout_web_vault.sh
+++ b/scripts/checkout_web_vault.sh
@@ -55,5 +55,5 @@ pushd "${VAULT_FOLDER}"
     # Update branch and tag metadata
     git fetch --depth 1 ${CHECKOUT_ARGS:-} vaultwarden "${VAULT_VERSION}"
     # Checkout the branch we want
-    git -c advice.detachedHead=false checkout FETCH_HEAD
+    git -c advice.detachedHead=false checkout "${VAULT_VERSION}"
 popd


tessus:
I think the problem is 2-fold.

Previously the process was clear. Download the repo at a specific commit, patch it, build it.

Now the repo is already pre-patched. There are many questions that are not explained in any documentation file or maybe I just missed it.

1. How is the web-vault pre-patched? There is no branch that holds all the changes that can be applied to a new branch at the commit from upstream. There are branches that are named like the version. So how is a new branch created. pull main to commit at new version upstream. switch into "old" branch (last pre-patched version), create new branch (new version to be released), rebase to main. Are those the steps to create a new rebased version?
2. how is the web-vault actually built? Are we still specifying a commit (which is the most recent commit in the version branch) or a branch? This can probably be determined by checking the workflow, but documentation would probably make things easier.

Both main questions create multiple subquestions. A few of them I have pointed out. Either way, as long as there is a reproducible process and the proper documentation, all should be as clear as it was before.

tessus:
Yep, it was by mistake. I clicked on the wrong email link. I thought I was in a different PR. project members can move a comment I think.

BlackDex (approved):

PR #201, opened 2025-05-28
use v2025.5.1 branch

stefan0xC:
Update to use new web-vault. Also added new css classes for the or text and login with passkey button and sso login button on the login page, cf. https://github.com/vaultwarden/vw_web_builds/commit/045b0cf85c429d76de424652376eacb3a564721f
So we can hide those elements explicitly and more reliably as suggested in https://github.com/dani-garcia/vaultwarden/pull/5890. (Once we update the web-vault to `v2025.5.1` we can change the selectors to `.vw-sso-login`, `.vw-passkey-login` and `.vw-or-text` respectively.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/vaultwarden/vw_web_builds/tree/v2025.5.0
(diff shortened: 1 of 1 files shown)


BlackDex:
Only when i actually create it. For which i didn't had time to fully start it.

BlackDex:
Yes a tag indeed. 
https://github.com/dani-garcia/bw_web_builds/blob/f5dbf839e4afa191aba7cefbeafd4b6489aa7502/.github/workflows/release.yml#L3-L6

BlackDex:
Well, i was working on other stuff. And merged this during a coffee break at my day job, and after that lost in other stuff. I'm a bit chaotic and multi-tasking all day long.

PR #206, opened 2025-06-30
update web-vault to v2025.6.2

stefan0xC:
The new web-vault `web-v2025.6.1` has done away with a feature flag that requires some changes in the Vaultwarden backend that we have not implemented yet. I.e. deleting entries would not be possible with newer clients ( https://github.com/dani-garcia/vaultwarden/discussions/5996). So I would wait for a new Vaultwarden release before updating to the newest web-vault.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/vaultwarden/vw_web_builds/tree/v2025.6.1
-ARG VAULT_VERSION=1b257e0cc69f766c2eec83d45e8be2274643ae98
+# Using https://github.com/vaultwarden/vw_web_builds/tree/v2025.6.2
+ARG VAULT_VERSION=483603bec7ce34ecea3cdf30d1b1757e2708d80c
 ENV VAULT_VERSION=$VAULT_VERSION
 ENV VAULT_FOLDER=bw_clients
 ENV CHECKOUT_TAGS=false
--- a/scripts/checkout_web_vault.sh
+++ b/scripts/checkout_web_vault.sh
@@ -2,7 +2,7 @@
 set -o pipefail -o errexit
 BASEDIR=$(RL=$(readlink -n "$0"); SP="${RL:-$0}"; dirname "$(cd "$(dirname "${SP}")"; pwd)/$(basename "${SP}")")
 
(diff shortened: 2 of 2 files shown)


tessus:
@stefan0xC I have noticed that in 2025.6.0, a few Send options are missing. e.g. expiration date, deactivate send, copy to clipboard.

I haven't been able to test 2025.6.2 yet. Has this been fixed?

Btw, you first comment mentions that vaultwarden has to be updated. The delete fix was already added, thus I was wondering, whether something else was missing from vw.

stefan0xC:
As far as I've looked into it this was intentionally added in `web-v2025.3.0`, cf. https://github.com/bitwarden/clients/commit/bfbad99fb733d8077aebd4da74beeddfd4602fdb so not sure what kind of fix you expect.

image

An my comment has not been updated since the fix was added. So it might be out of date.

stefan0xC:
Have you looked at the commit I've linked? That change was added in `web-v2025.3.0` that means it was already a part of that release. Not sure what explanation you want from me. It's not an error but seems intentional to me that this new dialog does not have the same options. If you think they are missing you should file a bug report upstream because we just make small changes to the web-vault.

tessus:
I am not saying that this is your fault. You clearly misunderstood my question:

I don't have a bitwarden account, thus I do not know what the Send dialog looks like now. I also said that I haven't tested 2025.6.2 yet. I thought the missing options were the result of a bug in the bitwarden web client, since these options are available on the desktop client.
Since you have seen (and most likely tested the 2025.6.2 web client), I thought that you would know, whether these options were re-added as part of a fix. (A fix in bw, not vw.)

As to your question, if I saw the commit you linked. Yes, I did. However, the bw commit messages and info in PRs are often useless, because they don't explain anything. The commit you sent mentioned that the send dialog was replaced. Well, great, but they didn't say that they removed 3 options and why. That's anyone's guess really.

I'll create a bw account and open an issue at bw.

stefan0xC:
> You clearly misunderstood my question

Why do you think I misunderstood your question? Which question are you referring to? Your initial question or the one where you were being intentionally obtuse?

> I thought the missing options were the result of a bug in the bitwarden web client, since these options are available on the desktop client.

And I answered that this was probably done intentionally, so there's no confusion on my part about your question. I'm not sure what you were hoping by questioning the semantics of my answer. I've referenced the commit where this was changed and heavily implied that this was not fixed because it does not look like a bug to me.

> I thought that you would know, whether these options were re-added as part of a fix

Like I said, this is not the case. btw: the screenshot was from the `v2025.6.2` release, which you could have also asked if that was not clear.

PR #213, opened 2025-08-08
update to v2025.7.2 branch

stefan0xC:
Seems like they have resolved the issue with the default theme (it now is using the System theme upstream too and the white flash is apparently gone). Our change https://github.com/vaultwarden/vw_web_builds/commit/c2b5d817c107a6c8e8236a8b16e1d5b6725ad810 still defaults to system in the preferences dialogue (not sure if this makes a difference).

I have not looked very much into the other changes yet.


BlackDex:
Cool. Yea i was using the 2025.7.1 version locally for a while already without any issues.
Ill go start using this one :)

stefan0xC:
I've updated the branch to include https://github.com/vaultwarden/vw_web_builds/pull/10

PR #214, opened 2025-08-20
update to v2025.8.0 branch

stefan0xC:
Update to newest [`web-v2025.8.0`](https://github.com/bitwarden/clients/releases/tag/web-v2025.8.0) release.

Our [patch set](https://github.com/bitwarden/clients/compare/web-v2025.8.0...vaultwarden:vw_web_builds:v2025.8.0) (temporarily) removes the integrations tab completely from the code (https://github.com/vaultwarden/vw_web_builds/commit/22f48fd3a062d6cda4f0a468055b738e857620f7) because it's dependent on proprietary code and would not compile otherwise and removing the offending code is easier to maintain than reverting the change and trying to keep our fork up to date with newer changes. (The dependency was most likely introduced in https://github.com/bitwarden/clients/pull/15757)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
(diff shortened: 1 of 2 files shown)


BlackDex:
I think integrations isn't part of OSS at all, so probably ok I think.

PR #216, opened 2025-09-30
update to v2025.9.0 branch and debian to trixie

stefan0xC:
update our web-vault to [web-v2025.9.0](https://github.com/bitwarden/clients/releases/tag/web-v2025.9.0)

also update the os of the container image to trixie

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@
 #    Note: you can use --build-arg to specify the version to build:
 #    docker build -t web_vault_build --build-arg VAULT_VERSION=main .
 
-FROM node:22-bookworm AS build
+FROM node:22-trixie AS build
 RUN node --version && npm --version
 
 # Can be a tag, release, but prefer a commit hash because it's not changeable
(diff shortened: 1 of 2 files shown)


tessus:
No sure, if you have seen it, but v2025.9.1 was released a couple of hours after you created the PR. I've set a notification for bw releases so I am always informed when a new version drops.
If you do something similar and/or want to not update it, please let me know. I'll never comment in a similar fashion again. I don't want to bother you with commwnts like that, but I currently genuinely don't know, whether you are aware of the new update.

tessus:
Ok, thanks for the info. In that case I won't bug you with "x is available" comments. ;-)

stefan0xC:
Updated to v2025.9.1.

I've decided to rework our patch that removes the seat limit check https://github.com/bitwarden/clients/commit/198e04864a1da376428a858c2963abdadc876466 a bit and this now also get's rid of the hint that says that you have 0 invites remaining because it was a bit misleading...

Old:
image

New:
image

PR #218, opened 2025-10-15
update to v2025.10.0 branch

stefan0xC:
Update to new web-vault release: [web-v2025.10.0](https://github.com/bitwarden/clients/releases/tag/web-v2025.10.0)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN node --version && npm --version
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
 # Using https://github.com/vaultwarden/vw_web_builds/tree/v2025.10.0
-ARG VAULT_VERSION=c6ed8bd4ff6426280764611c1773ee4a77ba442f
+ARG VAULT_VERSION=b5f5b2157b9b64b5813bc334a75a277d0377b5d3
 ENV VAULT_VERSION=$VAULT_VERSION
 ENV VAULT_FOLDER=bw_clients
 ENV CHECKOUT_TAGS=false


BlackDex:
I see several new endpoints by just logging in already. So those are probably something we need to either add a dummy for, or remove those calls from the web-vault it self. But it quickly looks like viewing at least works.

stefan0xC:
I think that those calls for getting warnings regarding free trail, reseller renewal and taxing can be disabled safely (cf. https://github.com/vaultwarden/vw_web_builds/commit/4422d7062857aa481ca1a15f2bee92730201aa54) in the web-vault (because I don't think those would affect other clients either), while the `/api/auth-requests/pending` endpoint should probably be added in Vaultwarden.

stefan0xC:
Looking at my changes to remove the warnings (https://github.com/vaultwarden/vw_web_builds/commit/4422d7062857aa481ca1a15f2bee92730201aa54 and https://github.com/vaultwarden/vw_web_builds/commit/7ac6d69336aadd86ab83f78907347a573ce5207a) I think that adding those warnings endpoints is actually way easier https://github.com/dani-garcia/vaultwarden/pull/6369 so we could probably undo the last two changes.

Timshel:
> It just stays on 3 seats

I believe it's probably because some logic is disabled elsewhere otherwise I believe it should diminish with each org user (At least that's how I understand [this](https://github.com/bitwarden/clients/blob/d65824e6241162e226ccc6b917aa353cc4d5781c/apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts#L329)).

As such I used U32::Max which is kind of ridiculous but at least there will always be some space ^^.

stefan0xC:
I've looked into it and it's capped at 20 so I've set it to that.

stefan0xC:
> I believe it's probably because some logic is disabled elsewhere otherwise I believe it should diminish with each org user (At least that's how I understand [this](https://github.com/bitwarden/clients/blob/d65824e6241162e226ccc6b917aa353cc4d5781c/apps/web/src/app/admin-console/organizations/members/components/member-dialog/member-dialog.component.ts#L329)).

Yeah, that's what I have believed to but I think that because we don't provide billing metadata (and possible because of our custom enterprise tier) it will always stay on 0.

Timshel:
@stefan0xC  

> I've looked into it and it's capped at 20

Seems low, I'm curious where did you see it ?

stefan0xC:
> Oh, yes the form is maxed at 20 but Org seats can probably have a way higher value.

I don't think that org seats are limited in Vaultwarden. And because that specific setting appears to not be used for anything else I don't see the point of setting it to a higher value? I rather feel like it would be bad UX if the dialog says you can invite 100 email addresses and when you submit more than 20 it says that you can only submit a maximum of 20.

stefan0xC:
reverted the changes (removal of the warnings and of the seat limit check) so it's easier to test the corresponding server-side PRs https://github.com/dani-garcia/vaultwarden/pull/6368 https://github.com/dani-garcia/vaultwarden/pull/6369 and https://github.com/dani-garcia/vaultwarden/pull/6371

Timshel:
Made the modification to allow to more easily test unreleased web-vault: https://github.com/dani-garcia/vaultwarden/pull/6380.

Did not finish fixing the tests in part due to the new extension spam :( (it's not only after creation ...). 
@stefan0xC do you intend to remove it ? I found it so obnoxious that I [did](https://github.com/Timshel/oidc_web_vault/commit/e16b5c62ef796e472bc63113c93fdeb9ad91ea0e).

stefan0xC:
I have adapted my login script for the test so it does not bother me. I don't think that we should dismiss it if that's just for the tests. But I have not really tested it extensively yet so I'm not sure how obnoxious it really is in practice. (But I'd suggest that you create a PR in the vaultwarden/vw_web_builds repo against the `v2025.10.0` branch, so we have a place to discuss your suggestion?

stefan0xC:
> I don't really understand why it's showing this page even if it detects that the extension is installed

Seems like they added this behavior intentionally https://github.com/bitwarden/clients/pull/16650

Issue #223, opened 2025-12-29
size of map files


tessus:
Sorry to open an issue, but discussions are disabled in this repo. I have noticed that there was a big size jump from version `v2025.10.1` to `v2025.12.0`.

By examining the files, it seems that the `.js.map` files are way bigger than before, even though the source files have not increased that much.
Was there a change in the tooling or the process of creating the application?

Afaik the map files are only used for debugging and not running the app, so the size won't affect the performance of the app itself, correct?

You can close this issue, since this is not a bug (or at least I don't think it is), but I would still like to use it as a discussion. I am curious why the size changed so drastically.

dani-garcia:
Yeah, that's fair. As far as I know the sourcemaps only get loaded when you first open the devtools in a browser and the rest of the time the browser doesn't even know they're there, so they shouldn't affect performance for the normal users.

tessus:
Perfect, in that case all is good and my worries are gone.

PR #224, opened 2025-12-30
fix organization creation

stefan0xC:
fixes https://github.com/dani-garcia/vaultwarden/issues/6638 by patching out https://github.com/bitwarden/clients/commit/12222e39b47cbeb10c87eb7a42c4f7490242f7d5

* skip validation of billing forms https://github.com/vaultwarden/vw_web_builds/commit/86538d369a4c48ef4eb4eea899527d49e7184da5 
* also remove unused videos to save some space https://github.com/vaultwarden/vw_web_builds/commit/0ed003259fb9c27b8a2be0e3c3c065c7cb4c26df as suggested by https://github.com/dani-garcia/bw_web_builds/issues/223#issuecomment-3697241697
* hide event logs link from the menu if events are not enabled https://github.com/vaultwarden/vw_web_builds/commit/6467e9b1fa88f9bf0206d4be564a1faa3b621ee2

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN node --version && npm --version
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
 # Using https://github.com/vaultwarden/vw_web_builds/tree/v2025.12.1
-ARG VAULT_VERSION=54af0ee1cc354cbe98e105c9d1bb9d8b45d293c1
+ARG VAULT_VERSION=6467e9b1fa88f9bf0206d4be564a1faa3b621ee2
 ENV VAULT_VERSION=$VAULT_VERSION
 ENV VAULT_FOLDER=bw_clients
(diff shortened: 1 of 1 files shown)


BlackDex:
Well, we haven't released a `a` or `b` version for a while either, so that might also still cause issues.
I'm not even sure if a patch digit breaks our stuff haha.

BlackDex:
Using a `+1` is also an option, that is valid metadata according to semver.

tessus:
Happy New Year.

The release was not triggered, b/c the regex no longer applies.

https://github.com/dani-garcia/bw_web_builds/blob/master/.github/workflows/release.yml#L7

should be at least `- 'v[0-9]+.[0-9]+.[0-9]+.*'` or better yet:

`- 'v[0-9]+\.[0-9]+\.[0-9]+(\.[0-9]+)?'`

I'll open a PR.

PR #226, opened 2025-12-31
ci: fix push trigger

tessus:
the current trigger no longer applies.

I've updated it accordingly. I also escaped the `.`, otherwise any character is valid.

e.g. `v1234h12l45` currently also triggers the workflow.

/ref https://github.com/dani-garcia/bw_web_builds/pull/224#issuecomment-3702994164

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,7 @@ permissions: {}
 on:
   push:
     tags:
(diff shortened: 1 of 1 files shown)


BlackDex:
> Hmm, ok. It's strange because I think I ran into a similar issue myself, but maybe I'm mixing it up. Also, i found it weird that even though a tag was created, the release or pre-release was missing.

The action was triggered, but i did not had time left to release it as a pre-release.

PR #232, opened 2026-02-11
Fix logo SVG viewBox clipping right edge of text

balcsida:
Small change, but this bugged me every time I wanted to log in on the web UI
Screenshot 2026-02-11 at 11 55 55


stefan0xC:
Since we don't use the resources from that repository anymore I don't think this change will do anything. Also I feel like there are some more other small glitches and alignment issues that are not present in the [original Inkscape SVG is in the main repository](https://github.com/dani-garcia/vaultwarden/blob/main/resources/vaultwarden-logo.svg) that should probably be fixed. For example this indentation:
image

balcsida:
> Since we don't use the resources from that repository anymore I don't think this change will do anything.

Could you please pont to where should I fix this then?

> Also I feel like there are some more other small glitches and alignment issues that are not present in the [original Inkscape SVG is in the main repository](https://github.com/dani-garcia/vaultwarden/blob/main/resources/vaultwarden-logo.svg?rgh-link-date=2026-02-11T11%3A18%3A39Z) that should probably be fixed. For example this indentation:

Maaan, I didn't notice that before but I cannot unsee that now... :DD
I'll fix it properly based on the original SVG then, but trying to apply some mindful SVGO magic on it.

stefan0xC:
The images are directly used in the web-vault repro, cf. https://github.com/vaultwarden/vw_web_builds/commit/450b1cab7f317dea4cdaca6ece8b696a640a995b
Some SVGs are also hard coded in typescript files (e.g. [bitwarden-logo.icon.ts](https://github.com/vaultwarden/vw_web_builds/blob/v2026.1.1/libs/assets/src/svg/svgs/bitwarden-logo.icon.ts)) so they would have to replaced there as well. I think I have fixed the rendering issues when I made the "Password Manager" and "Admin Console" variants but I have not gotten around to it to keep the changed images up to date with the ones in the resources folder (of either this or the main repository).

balcsida:
Thank you so much, I'll fix it there then
Also, thanks for bearing with me, this PR could have been an Issue 😅

balcsida:
I finally understood what you said, sorry for misunderstanding your comment. So this logo is basically cherry-picked to the vw_web_builds repo, I'm reopening my PR then

BlackDex:
Then you didn't understood it i think. This repo doesn't do anything anymore with those logos in this repo.
Nothing is using those.

I think it might be better to remove the resources here then? And locate it under the other repo?
Maybe even try to remove this repo at all maybe?

But, not sure though.

balcsida:
Damn, I'm so sorry then, again 😅 

How does this file get created, then?
https://github.com/vaultwarden/vw_web_builds/blob/v2026.6.2/apps/web/src/images/logo.svg

stefan0xC:
It's checked in as is (it was based initially on the resources here but like I said some of those files diverged because I only updated them in the `vaultwarden/vw_web_builds` repo). I would recommend having a definitive source of truth (e.g. the [resources folder](https://github.com/dani-garcia/vaultwarden/tree/main/resources) in the main vaultwarden repo) and then I can update the rebranding commit (e.g. https://github.com/vaultwarden/vw_web_builds/commit/bb6888b3b9fc97e0150b3faf364ff576db306907) in our patch set from there.

BlackDex:
Well, lets try to move all these images into the vaultwarden repo then to keep it centralized and a single source of truth for all these kind of resources.

Therefor going to close this one.

PR #234, opened 2026-02-11
fix light mode color scheme of side panel

stefan0xC:
fix oversight of change of the css color scheme in `web-v2026.1.1` by turning the side bar back to dark gray:
https://github.com/vaultwarden/vw_web_builds/commit/0a2cd9d14281306a98d627573edebb8be5fd3688

PR #236, opened 2026-02-19
update to v2026.2.0 branch

stefan0xC:
update to new web-vault release: [web-v2026.2.0](https://github.com/bitwarden/clients/releases/tag/web-v2026.2.0)

this now includes the css change directly so #234 can be closed if we don't want to make a hotfix release.


BlackDex:
If this works without issues, then I'd rather strait go to this version.

stefan0xC:
As far as I've tested it it works but I have not looked for new features or checked for error messages very thoroughly.

BlackDex:
Other than a small UI change regarding sends, I'm not seeing anything changed or broken either.

PR #238, opened 2026-03-04
update to v2026.3.1 branch

stefan0xC:
update to new web-vault release: [web-v2026.2.1](https://github.com/bitwarden/clients/releases/tag/web-v2026.2.1)

I think some server side changes are probably necessary for email protected Sends to work but I have not really looked into it.

I've also added the css class `vw-remember` https://github.com/vaultwarden/vw_web_builds/commit/84903f6dc0fe59fb873170381f7949e6ffd010e8 so we can hide the checkbox position independently, cf. https://github.com/dani-garcia/vaultwarden/pull/6852

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.2.1
-ARG VAULT_VERSION=280c5d4285faee673c4ff044ad35110c7e3cde05
+# Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.3.1
+ARG VAULT_VERSION=cd1eb788b667bdf31aee2d8b32aae7718c59097e
 ENV VAULT_VERSION=$VAULT_VERSION
 ENV VAULT_FOLDER=bw_clients
 ENV CHECKOUT_TAGS=false
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.2.0
-ARG VAULT_VERSION=3986035910f6ad172c8ceabcd378e2ca0d25349c
+# Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.2.1
+ARG VAULT_VERSION=280c5d4285faee673c4ff044ad35110c7e3cde05
 ENV VAULT_VERSION=$VAULT_VERSION
 ENV VAULT_FOLDER=bw_clients
 ENV CHECKOUT_TAGS=false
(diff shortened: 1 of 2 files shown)


stefan0xC:
seems like the path of the requested billing metadata has been changed. I think [the old path](https://github.com/dani-garcia/vaultwarden/blob/ba5519167634ebe1e1f0fc10d610d10d1f405101/src/api/core/organizations.rs#L2146) can be deprecated?
```rust
#[get("/organizations/<_org_id>/billing/vnext/self-host/metadata")]
fn get_billing_metadata(_org_id: OrganizationId, _headers: Headers) -> Json<Value> {
    // Prevent a 404 error, which also causes Javascript errors.
    Json(_empty_data_json())
}

```

stefan0xC:
also the `X-Request-Email` seems to be (possibly) padded `BASE64URL` and [not `BASE64URL_NOPAD` anymore](https://github.com/dani-garcia/vaultwarden/blob/ba5519167634ebe1e1f0fc10d610d10d1f405101/src/api/core/accounts.rs#L1331).

BlackDex:
> also the `X-Request-Email` seems to be (possibly) padded `BASE64URL` and [not `BASE64URL_NOPAD` anymore](https://github.com/dani-garcia/vaultwarden/blob/ba5519167634ebe1e1f0fc10d610d10d1f405101/src/api/core/accounts.rs#L1331).

Looks like it indeed: https://github.com/dani-garcia/vaultwarden/issues/6960
I already think i know how to fix this in a good way to still support the old format too.

stefan0xC:
Creating a folder also does not work at the moment, cf. https://github.com/dani-garcia/vaultwarden/issues/6962

BlackDex:
> Creating a folder also does not work at the moment, cf. [dani-garcia/vaultwarden#6962](https://github.com/dani-garcia/vaultwarden/issues/6962)

Which is weird, since i am able via 2026.2.1 Android version.

stefan0xC:
Sorry, I meant creating an item when explicitly selecting the "no folder" folder. The new web-vault also sends the `folderId: ""` instead of `folderId: null` after selecting the option.

BlackDex:
So, I wonder, is it still good to check/merge this? Or should we try to go strait to v2026.4.1?
Or go to v2026.3.1 first, adjust the server where needed, since the April release has a lot of changes and new/adjusted features.

stefan0xC:
I have the `web-v2026.3.1` ready. The new releases I have not checked yet, it seems like the `web-v2026.4.1` release cannot be built because of an out of sync `package-lock.json`. (I can update it manually but not sure if we really want that...)

edit: seems like this is caused by using a newer version of npm. with `v22` and `v23` it should build just fine.

stefan0xC:
hm... the link to the Subscription page is visible again.
update: should be fixed via https://github.com/vaultwarden/vw_web_builds/commit/cd1eb788b667bdf31aee2d8b32aae7718c59097e

BlackDex:
Ok, until now, i have found one new endpoint, but that was an easy fix.
I also found a way to hide the FireFox and DuckDuckGo email forwarders 🥳 .

I'm still checking, but not much time today.

stefan0xC:
Looks like `web-v2026.4.1` has at least one changed endpoint as well ( `/identity/accounts/prelogin/password` instead of `/identity/accounts/prelogin`). I have not really checked the newly removed feature flags (well, the Archive button is visible but there's a PR already that I have not tested and also the Admin Console has been cleaned up a bit - especially the Policies tab):
image

BlackDex:
Ah, the policies look nicer now.

BlackDex (approved):

PR #243, opened 2026-04-22
update to v2026.4.1 branch

stefan0xC:
update to new web-vault release: [`web-v2026.4.1`](https://github.com/bitwarden/clients/releases/tag/web-v2026.4.1)

as said in #238 the `/identity/accounts/prelogin` has been changed to `/identity/accounts/prelogin/password` and also some feature flags have been removed already in [`web-v2026.4.0` ](https://github.com/bitwarden/clients/releases/tag/web-v2026.4.0)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.3.1
-ARG VAULT_VERSION=cd1eb788b667bdf31aee2d8b32aae7718c59097e
+# Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.4.1
+ARG VAULT_VERSION=e564bf7c687fb12a88dd8b53beff726eb23ea932
 ENV VAULT_VERSION=$VAULT_VERSION
 ENV VAULT_FOLDER=bw_clients
 ENV CHECKOUT_TAGS=false
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ RUN node --version && npm --version
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
 # Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.4.1
-ARG VAULT_VERSION=e564bf7c687fb12a88dd8b53beff726eb23ea932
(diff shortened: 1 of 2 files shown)


tessus:
I have noticed in the 2026.3.1 the following 2 lines that look a bit strange:

image

I think in the bw web vault they are used to separate news feed and the upgrade plan button with the rest of the sidebar. Is there any chance to get rid of them in vw-web?

tessus:
Oh, I think I was wrong. It's for the `Password Manager` and `Admin Console` entries, which do not show up when the user does not belong to any orgs. So in that case it looks a bit weird, but I guess there's nothing one can do...

stefan0xC:
I think Bitwarden also uses the space to display advertisement? I'll have to check how it looks without an organization and if there is something that could be done about it

stefan0xC:
As far as I've looked those two `<bit-nav-divider>` elements are hardcoded into the sidebar before [the product-switcher](https://github.com/vaultwarden/vw_web_builds/blob/e6e821395b14bc279978cb42957ebe3a68fe936c/apps/web/src/app/layouts/web-side-nav.component.html) and [before the toggle button](https://github.com/vaultwarden/vw_web_builds/blob/e6e821395b14bc279978cb42957ebe3a68fe936c/libs/components/src/navigation/side-nav.component.html#L37). Perhaps we could move the first one to the [navigation-switcher.component.html](https://github.com/bitwarden/clients/blob/b9658a795ff0aa10f927667f8b286c4ec1552d9a/apps/web/src/app/layouts/product-switcher/navigation-switcher/navigation-switcher.component.html#L2) so it will only be shown if there are also items to be displayed? I don't think that this change would be harder to maintain than adding custom classes for CSS.

update: the straight forward change that i did https://github.com/vaultwarden/vw_web_builds/commit/876c4dbd526add221223d4869d75eb61efe57e37 should have no unwarranted side effect since we never show more products or have the upgrade plan button in Vaultwarden

BlackDex (approved):

tessus:
In the light theme the horizontal bars are now blue instead of grey:

image

Shall I open a new issue?

stefan0xC:
Ah, thanks for pointing it out. Sorry, I have not checked light mode. I'll fix it in the next release, so a separate issue is not needed. A quickfix using [custom css](https://github.com/dani-garcia/vaultwarden/wiki/Customize-Vaultwarden-CSS) would be to set it to the same as dark mode:

```css
:root {
    --color-nav-border-primary: var(--color-gray-700);
}
```

image

tessus:
no worries, it's not really important. just wanted to mention it.

tessus:
oh, btw I also use a stylus user style, so that notes are shown in a monospace font. ;-)
maybe I should use custom css instead 🤔

PR #244, opened 2026-05-01
 skip email step when SSO is the only auth method

IacopoSb:
New patch `patches/v2026.4.1.patch` for the v2026.4.1 web-vault that improves the login UX when `SSO_ONLY=true` is set on the vaultwarden side.

## Why
Today, with SSO_ONLY enabled, the login page still shows the email input field and the "Remember email" checkbox, and clicking the SSO button requires entering an email even though that email is ignored by the backend (single-tenant vaultwarden uses `FAKE_SSO_IDENTIFIER` server-side).

Two small changes were made in `libs/auth/src/angular/login/login.component.{html,ts}`:
- `class="vw-email-form-field"` on the `<bit-form-field>` wrapping the email inputs, and `class="vw-remember-email"` on the Remember-email `<bit-form-control>`. These hooks let the vaultwarden SCSS template hide both elements when `sso_only=true`.
- `handleSsoClick()` now tolerates an empty email: if the form has no email it skips validation and proceeds. Single-tenant vaultwarden does not require an email to start the SSO flow.

## Test done
- Built locally with `make full`. With `SSO_ENABLED=true SSO_ONLY=true` and the companion vaultwarden SCSS change, the login page shows only the "Continue with SSO" button.
Screenshot 2026-05-01 163036

### DIFF
--- a/patches/v2026.4.1.patch
+++ b/patches/v2026.4.1.patch
@@ -1,0 +1,53 @@
+diff --git a/libs/auth/src/angular/login/login.component.html b/libs/auth/src/angular/login/login.component.html
+index f7240601..ebc57dcf 100644
+--- a/libs/auth/src/angular/login/login.component.html
++++ b/libs/auth/src/angular/login/login.component.html
+@@ -13,7 +13,7 @@
+ <form [bitSubmit]="submit" [formGroup]="formGroup">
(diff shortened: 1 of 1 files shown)


BlackDex:
This is the wrong location for something like this.
Either this needs to be done dynamically via the Vaultwarden server CSS, or somewhere in the vw_web_build repo.
Therefor going to close this here.

PR #245, opened 2026-05-29
update to v2026.6.2 branch

stefan0xC:
update to new web-vault: [`web-v2026.5.0`](https://github.com/bitwarden/clients/releases/tag/web-v2026.5.0)

I've not really looked very deeply into the list of changes but it seems like they removed at least the feature flag `pm-27086-update-authentication-apis-for-input-password`  https://github.com/bitwarden/clients/pull/20306 that requires us to change a few things.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.6.0
-ARG VAULT_VERSION=13f9a3ae689f422b9c5ef3bb63c7c8ddb5035f45
+# Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.6.2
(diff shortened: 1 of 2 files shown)


stefan0xC:
btw: I was testing if https://github.com/vaultwarden/vw_web_builds/commit/fa4722a72b5ef8a113e1f398d2306aa56e31280d would break things but it seems like it still works for each of the three cases (!sso_enabled, sso_enabled && !sso_only , sso_enabled && sso_only) so maybe this does nothing? 
edit: removed again  `v2026.6.0`

stefan0xC:
update to [web-v2026.6.0](https://github.com/bitwarden/clients/releases/tag/web-v2026.6.0)

stefan0xC:
update to [web-v2026.6.2](https://github.com/bitwarden/clients/releases/tag/web-v2026.6.2)

@Timshel I have not tested all your pull requests yet but I hope that they are still compatible.  btw: As far as I've tested it I had to adapt my playwright tests because the required fields now contain a `*` and some other small changes like that.

Timshel:
@stefan0xC just tested a `v2026.6.2` based build and as you mention outside of the field name change and a fix with the server config everything appears to work.

BlackDex (approved):

PR #249, opened 2026-07-24
update to v2026.7.0 branch

stefan0xC:
update to new web-vault: [web-v2026.7.0](https://github.com/bitwarden/clients/releases/tag/web-v2026.7.0)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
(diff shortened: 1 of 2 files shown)


BlackDex:
Nice. I think it will probably work just fine after some pending PR's are merged. But lets first release with v2026.6.4.
Will see if i have some time this weekend to validate this better.

tessus:
would also require (nice to have)

- https://github.com/dani-garcia/vaultwarden/pull/7478
- https://github.com/dani-garcia/vaultwarden/pull/7502

BlackDex:
As far as i have quickly tested it, it works fine.

Those PR's are nice to add, but are definitely not required, without those feature flags the web-vault works just fine, just not showing those new types.

BlackDex (approved):

PR #250, opened 2026-08-05
update to v2026.8.0 branch

stefan0xC:
update to new web-vault: [web-v2026.7.1](https://github.com/bitwarden/clients/releases/tag/web-v2026.7.1)

The feature flag `pm-34145-policies-in-accepted-state` has been removed (https://github.com/bitwarden/clients/pull/21761) so it requires the policies to have a revision date as well. Otherwise if you are member of an organization with a policies [this call will fail](https://github.com/bitwarden/clients/blob/7b73ec614dfcfb192e5552cdf0fb5380d784746d/libs/common/src/admin-console/models/domain/policy.ts#L67) and this exception will make the web-vault completely inaccessible.
image

So this should be addressed first in the Vaultwarden backend. (I can do a PR for returning an empty revision date to make it compatible (i.e. add `"revisionDate": null,`) but I currently don't have the time to look properly into this.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN node --version && npm --version
 # Can be a tag, release, but prefer a commit hash because it's not changeable
 # https://github.com/bitwarden/clients/commit/${VAULT_VERSION}
 #
-# Using https://github.com/vaultwarden/vw_web_builds/tree/v2026.7.0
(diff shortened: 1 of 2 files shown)


stefan0xC:
update to the latest release [`web-v2026.8.0`](https://github.com/bitwarden/clients/releases/tag/web-v2026.8.0)

